Issue #17, opened 2020-08-21
Change The Title Bar contents


AbdulrhmnGhanem:
Title bar contents should vary for authenticated and anonymous users.
- Add project button should only be visible to authenticated users.
- Shall we group `chat`, `Twitter`, and `GitHub` in an `about` route?
- A login/logout button.

PR #18, opened 2020-08-22
Add User accounts

### LINKED ISSUE #6: Implement login
We'll want to hide away most of gitea and allow logging in via our own frontend. Ideally we can also enable "login with github/facebook/google".

AbdulrhmnGhanem:
Closes #6.


kasbah:
Let me know when you think this is ready to be merged @AbdulrhmnGhanem

frontend/components/SignUpForm.js
```diff
@@ -1,14 +1,26 @@
...
 import { SignUpForm } from '../models/SignUpForm'
 
+const endpoint = `${process.env.KITSPACE_GITEA_URL}/user/kitspace/sign_up`
+
+
 export default function () {
   const [form, onChange, isValid, errors] = useForm(SignUpForm)
 
-  const submit = () => {}
+  const submit = async () => {
```
AbdulrhmnGhanem, line 15: @kasbah  
- It result in cors error.
- Access to XMLHttpRequest at 'http://gitea.kitspace.test:3000/user/kitspace/sign_up' from origin 'http://kitspace.test:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
- Aren't the Nginx configuration suppose to solve this? I know I am missing something.

kasbah: It hasn't been enabled for this case / endpoint I believe. We should be able to solve it with either nginx config or within gitea.

kasbah: Basically we have to allow the apex domain  origin (`kitspace.test`) on these endpoints somehow.

kasbah:
I just tried the UI using the branch on #24. Overall, it's really good! It's really good to see you have gone through all the possible routes and edge cases I could think of and it looks really good. Some minor stuff about the UI itself:

- the input moves when the message appears, this makes it a bit annoying to use
- how will oauth buttons look if there are more than three?
- can we make oauth buttons look more official?
- might look better without the icons in the password, email, user inputs
- url should be /log_in and /sign_up not /auth
- title should be "Log in" not "Login"
- "Password" is capitalized, the rest aren't

AbdulrhmnGhanem:
> * the input moves when the message appears, this makes it a bit annoying to use

Do you want the message to appear at the bottom or just put a place holder for it so the form doesn't move?


> * how will OAuth buttons look if there are more than three

For Four we might make the form wider, more than four(I think this will be overkill) we can make them stacked vertically.

> * can we make OAuth buttons look more official?

Yeah, then I'll write custom buttons instead of using the ones provided with Semantic-UI


> * url should be /log_in and /sign_up not /aut

Does this mean I have to separate each form to a page by itself? 



> * might look better without the icons in the password, email, user inputs
> * title should be "Log in" not "Login"
> * "Password" is capitalized, the rest aren't

OK, easy fixes.

kasbah:
> Do you want the message to appear at the bottom or just put a place holder for it so the form doesn't move?
I don't have a specific solution in mind. 

> Does this mean I have to separate each form to a page by itself?
Nah, just rename it to /login[.js] for now

Another easy fix:

- Needs more spacing at the top between the title bar and the form

kasbah:
I can't reproduce that. I can remove the `add_header` directives fine and it re-starts: https://github.com/kitspace/kitspace-v2/commit/dccaec660857a3229df326d1141648be117ec496

AbdulrhmnGhanem:
And here we go! :rocket:
I didn't test it extensively though I'll do that tomorrow. If I found enough time, I'll write tests.

AbdulrhmnGhanem:
@kasbah, I need your input on these points so I can finish:

 How are we going to handle OAuth?
    - Gitea doesn't provide a way to authorize users from external OAuth providers. Gitea itself can work as an OAuth provider, not the other way around. The closest thing to OAuth is the [OpenID](https://github.com/kitspace/gitea/blob/bfb404274f4a48e7c058ffabcb9f103d161f8594/routers/user/auth_openid.go). I can work on something similar.
- How are we going to handle authorized pages?
    - The pages should be behind the authorization wall.
- I'm going to finish the navbar with these considerations
    - When the user is signed in the navbar will have `Add a project` and `sing out` buttons.
    - When the user is signed out the navbar will have `Make contact` and `sign in` buttons.
    - When a user completes a login successfully he will be redirected to the page he pressed on the sign-in form.
    - If an already logged in user goes to `/login`, how is this case going to be handled?
- Did you come up with an idea for preventing the layout shifts when the messages appear?

I think we can merge the current PR and let the OAuth work for a separate one, in this case, I will write some tests so we can merge confidently. Do you have a preferred testing library? I think [Cypress](https://www.cypress.io/) is fine.

kasbah:
Hey, 

- I believe "Open ID" refers to "Open ID Connect" which is an extension of OAuth. In any case, using that is fine as long as that's what allows us to do "login with twitter", "login with github" etc.
- We should be able to see if a user is logged in in NextJS and serve pages accordingly. I guess you can call that
- I think we should keep "Add a project" even when users are not logged in, it's a good general call to action for all people. Basically just leave the nav bar as is but add add a "Log In/Sign up" or user-profile menu button (the meny can just contain "Sign out" for now) depending on weather the user is signed in or out. Re-direct them to log in as when they click "Add a project"
- Layout shifts, can you not solve it somehow? I'll take a look if you are really out of ideas but you are doing a good job with the styling so far and I am sure you can think of a way not to make the layout shift. 

Give me a while to test and review, can you resolve the merge conflicts first of all?

kasbah:
I just tested it out. Seems to work. One thing I noticed is that I think it should accept either a username or an email, that's how it is on Gitea anyway.

AbdulrhmnGhanem:
>  One thing I noticed is that I think it should accept either a username or an email, that's how it is on Gitea anyway.

Yeah, it's a matter of changing the label! :laughing:

So I'll write the tests and you can merge this branch tomorrow morning. I'll also start working on the OAuth.

AbdulrhmnGhanem:
- The tests have some duplication that can be extracted into tasks.
- `Superagent` isn't working well with `Cypress`.
- At least now we have some kind of safety net which can be improved as we go.

I also update the readme with instructions to run the tests.

AbdulrhmnGhanem:
Whenever https://github.com/MichalSzorad/react-social-login-buttons/pull/46 get merged, we will no longer need `components/GoogleLoginButton`.

AbdulrhmnGhanem:
I think this is ready to be merged.
- For Navbar, it's tracked at #17 
- For layout shifts, it's tracked at #25

frontend/cypress.example.json
```diff
@@ -0,0 +1,7 @@
+{
+  "baseUrl": "http://kitspace.test:3000",
+  "env": {
+    "gitea_admin_username": "your gitea admin username",
+    "gitea_admin_password": "your gitea admin password"
+  }
+}
```
kasbah, line 7: Can we grab this from the .env ?

nginx/kitspace.template
```diff
@@ -46,16 +46,18 @@ server {
...
+		proxy_set_header           X-Forwarded-For                    $proxy_add_x_forwarded_for;
+		proxy_set_header           X-Forwarded-Proto                  $scheme;
+		add_header                 'Access-Control-Allow-Origin'      '@KITSPACE_URL' always;
+		add_header                 'Access-Control-Allow-Methods'     'GET, POST, PUT, PATCH, DELETE, OPTIONS';
+		add_header                 'Access-Control-Allow-Credentials' 'true';
+		add_header                 'Access-Control-Allow-Headers'     'cache-control, pragma, user-agent';
+		proxy_cookie_path          ~^(.+)$ "$1;                       domain=.@KITSPACE_DOMAIN";
+		proxy_set_header           Origin                             gitea.@KITSPACE_DOMAIN;
+        proxy_pass_request_headers on;
+
```
kasbah, line 60: indentation needs to be fixed here

frontend/pages/newsletter.js
```diff
@@ -0,0 +1,82 @@
...
+            />
+          </Form.Field>
+          <Button primary type="submit" style={{ marginTop: 30 }}>
+            Subscribe
+          </Button>
+        </Form>
+      </Container>
+    </div>
+  )
+}
```
kasbah, line 82: As mentioned before, no need to copy this over.

kasbah:
I think you should finish the nav bar at least as part of this PR.

frontend/models/SignUpForm.js
```diff
@@ -0,0 +1,9 @@
+import Joi from 'joi'
+
+export const SignUpForm = Joi.object({
+  username: Joi.string().alphanum().required().max(20),
+  email: Joi.string()
+    .email({ tlds: { allow: false } })
+    .required(),
+  password: Joi.string().pattern(new RegExp('^[a-zA-Z0-9]')).min(6).required(),
+})
```
kasbah, line 9: So, are we sure this validation is all exactly as we set it up in Gitea?

frontend/cypress.example.json
AbdulrhmnGhanem: No, it isn't OS env https://docs.cypress.io/guides/guides/environment-variables.html#Setting

frontend/models/SignUpForm.js
AbdulrhmnGhanem: Not sure, if you added additional constraints to the password as mentioned in #21, this will need to be updated. Also don't know the max limit for username. With that said that if it passed this validation Gitea will accept it.

frontend/cypress.example.json
AbdulrhmnGhanem: Oh, I think that is possible https://docs.cypress.io/guides/guides/environment-variables.html#Option-3-CYPRESS

AbdulrhmnGhanem: Done!

kasbah:
If you don't want to finish of the nav bar as part of this PR then you should at least restore it to how it was. It seems like it's just broken currently.

frontend/components/TitleBar.js
```diff
@@ -1,115 +1,159 @@
...
+function SigningButton(props) {
+  const isAuthenticated = props.auth
+  return (
+    <Menu.Item>
+      <Button color={isAuthenticated ? 'red' : 'green'} href="#">
+        {isAuthenticated ? 'Sign out' : 'Sign in'}
+      </Button>
+    </Menu.Item>
+  )
+}
```
kasbah, line 159: Just undo all of this?

kasbah (changes requested):

frontend/models/SignUpForm.js
kasbah: Is there anyway we can unify this? Maybe not for this PR but in the future? Would be good to just define our rules in one place...

AbdulrhmnGhanem: Yes. Though I have thought of it, I think it's an antipattern; usually, models are stand-alone modules except for composite models. Also, no one validates the password for login forms. I had the same feeling though I didn't manage to come with a strong argument to make it a legit decision. Assuming the worst-case scenario the API will respond with the right error message.
I think the consistency should be considered on the testing level, otherwise, it's like writing React with Angular flavor!

kasbah:
I just tried it again and noticed some issues:

- It doesn't seem to actually log me in! If I go back to the main page it still says "hi there unknown user" even when i do a refresh
- Password requirements don't seem right, it's not letting me use a password with special characters
- Password message shows the password but it should be hidden

frontend/models/SignUpForm.js
kasbah: I am not sure I understand what you are proposing. What do you suggest we do?

kasbah (approved):

README.md
```diff
@@ -42,3 +42,18 @@ docker-compose up
 5. Go to [gitea.kitspace.test:3000/install](http://gitea.kitspace.test:3000/install) and complete the install (everything should already be filled in correctly). Create a new user and login.
 
 6. Making edits on the code in `frontend/` should auto compile and hot-reload at [kitspace.test:3000](http://kitspace.test:3000).
+    
+7. If you add a dependency to `frontent/package.json`, rebuild the frontend image using `docker-compose --no-cache frontend`.
```
kasbah, line 46: `docker-compose` is missing `build` I think.

AbdulrhmnGhanem:
> * Password requirements don't seem right, it's not letting me use a password with special characters
> * Password message shows the password but it should be hidden

Fixed!



> * It doesn't seem to actually log me in! If I go back to the main page it still says "hi there unknown user" even when i do a refresh

It does, one of the e2e tests checks the database to see the last login. The message in the index uses `session` which isn't being passed by Gitea from the first place - AFAIR, I rebuilding Gitea to see. Passing the user and its state was what I was working on before breaking the kernel

kasbah:
What do you mean by `session`? The cookie? Why is not being passed? If I login via gitea.kitspace.test it works fine.

AbdulrhmnGhanem:
I'm not sure how the current index knows the user I thought you were passing a session to the frontend then found out it's just the window session. But what I am sure of is the user gets logged in Gitea and `i_like_gitea` gets set.

kasbah:
`i_like_gitea` seems to get set when we request any page via Gitea, (i.e. any page as the frontend is proxied through `gitea.kitspace.test:3000/__kitspace` ). It seems like Gitea doesn't think we login when using your login page, whereas if I login via gitea it's fine. `i_like_gitea` doesn't seem to change at all. 

Can you figure out how to make it work? What it does currently isn't very useful. :smile:

AbdulrhmnGhanem:
Knowledge duplication once again. :sweat_smile: 
* In the `index` page you access the user by accessing a session object, [here](https://github.com/kitspace/kitspace-v2/blob/ac38476b8baf180601695b3cabe662b79ac2cff6/frontend/pages/index.js#L67). When I asked you about it you didn't get me!
* The `session` is being set [here](https://github.com/kitspace/gitea/blob/bfb404274f4a48e7c058ffabcb9f103d161f8594/routers/kitspace.go#L40), which has a user field. So in this case it makes sense that `user` can be accessed by the `index` page.
    * You did this work before I jumped in and till this moment I didn't even know it existed! :sweat_smile: 
* For the implementation in my head, I was taking a completely different approach.
    * I am working on it in #26.
    * In this case, we won't need `getIntialProps` which means we can SSR it. Also, it will work with protected routes.

kasbah:
- #26 seems to be about frontend changes, I can't see what your approach would be yet.
- I don't understand... using `getInitialProps` is what makes SSR work!
- I think the issue is that through this new endpoint Gitea doesn't think we are logged in. Otherwise this existing method would work as it does if I login through `gitea.kitspace.test`

(I did try and give a [high level explanation of what's going on in the existing implementation before](https://github.com/kitspace/kitspace-v2/issues/6#issuecomment-657588851).)

AbdulrhmnGhanem:
I meant prerendering at the build time with `getStaticProps`. Anyways this will be beneficial with repo pages more the index.
I'll look into Gitea again to see what I can do.

kasbah:
@AbdulrhmnGhanem turns out the issue was we were not sending cookies in the `fetch` requests. It works now, you still have to do a full refresh to see that the user is logged in.

AbdulrhmnGhanem:
So the login is ok now. We have to finish the OAuth providers then, On the
staging site, you showed me there's a google login button which doesn't
exist on my local version. Did you configure Gitea in a special way to get
it to work?

kasbah:
You mean Github, yeah? It's set up in https://gitea.staging.kitspace.org/admin/auths. I am not sure if it needs a public webhook endpoint though or if you can do it on localhost. 

I'll set you up with a dev server anyway. Send me your public key on Element.

AbdulrhmnGhanem:
What was the design you had in mind for authenticated routes? I know you it's solved in the index page through accessing the cookies but this means each page will need `getIntialProps` to access the cookies. Did you think of centralizing it(the current authentication state)? Do I miss something? I tried to make an endpoint for gitea that will be called on each render to get the authentication state, though I found myself duplicating what you have already done in the `kitspace` route. I am asking to be on the same page before implementing something.

kasbah:
I think we can put the logic that is in `index.js` into `_document.js`. Note that in `index.js` we save the session onto the `window` object and can retrieve it at any time using the [`getSession` function](https://github.com/kitspace/kitspace-v2/blob/ac38476b8baf180601695b3cabe662b79ac2cff6/frontend/pages/index.js#L25-L32). No custom endpoint needed. 

Happy to jump on a call to discuss if it helps.

AbdulrhmnGhanem:
I'll try this. Let's see if I can get by tomorrow. Otherwise we would make
the call!

kasbah:
We'll need to protect against CSRF on these endpoints.

kasbah:
We need a remember me box.

AbdulrhmnGhanem:
I am getting some weird error after submitting the sign-in form `Nginx 400: bad request`. I am skeptical of either I used the wrong version to build the docker network or trying to sign in while being signed in.

AbdulrhmnGhanem:
> I guess I should mention that it used to set `window.session = session` at some point. It doesn't appear to do that at the moment, but that was the intention.

It's set in `_app.js`

frontend/components/Page.js
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export const Page = props => {
+  return (
+    <AuthProvider>
+      <Head
+        description={props.head?.description}
+        ogImage={props.head?.ogImage}
+        title={props.head?.title}
+        url={props.head?.url}
```
kasbah, line 44: I think this is worse interface than before. 

1. Using an object as a prop is a bad practice since you can change a field and it will still be seen as equal
2. Being able to specify things on a page using `<Head>` seems much nicer than passing in an object to our `<Page>` as a prop.

AbdulrhmnGhanem: 1. Using an object as a prop is a bad practice since you can change a field and it will still be seen as equal

Shouldn't the head content change between different renders of the same page?!

> 2. Being able to specify things on a page using `<Head>` seems much nicer than passing in an object to our `<Page>` as a prop.

Ok, We can remove it from the `Page` and add it as a child component like other components inside `Page`.

kasbah: It's not really an issue in this case, as it is, but when you are designing interfaces on React components we should just avoid having `object`  (including `Array`) as a prop. The reasoning is that they are mutable but will continue to pass equality with `===`  when changed. When you try and optimize components with these mutable `object`s as props, it's a pain. 

> Ok, We can remove it from the Page and add it as a child component like other components inside Page.

Great, yeah. Is it actually possible to use `Head` multiple times in the same page so we can still have some defaults?

frontend/pages/index.js
```diff
@@ -1,20 +1,16 @@
...
-      <TitleBar route="/" />
+    <Page title="home">
       <div>
         Hi there {name}, {_csrf}
         <pre>{JSON.stringify(repos, null, 2)}</pre>
       </div>
-    </>
+    </Page>
   )
 }
```
kasbah, line 16: I'm guessing `getSession`, below,  isn't used anymore and should be removed.

kasbah:
Since this is now merged into `poc-shop` and deployed on https://shop.kitspace.org I've noticed that the page loading spinner flashes, seemingly uneccessarily. I'd like to get rid of it. What is it actually loading? I'd much prefer to show the skeleton and text on the page as quickly as possible and add progressively add anything that takes longer to load.

frontend/components/NavBar.js
```diff
@@ -0,0 +1,239 @@
...
+  return (
+    <div className={styles.titleBar}>
+      <BigBar isProjectRoute={isProjectRoute} isSubmitRoute={isSubmitRoute} />
+      <SmallBar isProjectRoute={isProjectRoute} isSubmitRoute={isSubmitRoute} />
+    </div>
+  )
+}
+
+function BigBar({ isProjectRoute, isSubmitRoute }) {
+  /* This is the Navbar render on big screens */
```
kasbah, line 26: render*ed*,

frontend/components/Page.js
```diff
@@ -0,0 +1,66 @@
...
+  if (loading) {
+    return (
+      <Dimmer active inverted>
+        <Loader>Loading...</Loader>
+      </Dimmer>
+    )
+  } else {
+    return <Container style={{ marginTop: 30 }}>{children}</Container>
+  }
+}
```
kasbah, line 35: Just found out this is the cause of the loading spinner. We need a better solution for this.

frontend/components/SignUpForm.js
```diff
@@ -0,0 +1,110 @@
...
+      const { error, message } = data
+      setApiResponse({
+        error: error || 'API error',
+        message: message || 'Something went wrong. Please, try again later.',
+      })
+    }
+  }
+
+  const hasApiError = apiResponse.error !== undefined
+  const isSuccessfulReg = apiResponse.duration !== undefined
```
kasbah, line 36: Generally use `== null` or `!= null` for null/undefined checks please.

frontend/cypress/support/commands.js
```diff
@@ -0,0 +1,94 @@
+const signUpEndpoint = 'http://gitea.kitspace.test:3000/user/kitspace/sign_up'
```
kasbah, line 1: what does the folder name `support` indicate? should these tests maybe be called `gitea-api` or something like that?

frontend/hooks/useForm.js
```diff
@@ -0,0 +1,30 @@
...
+    e.persist()
+    setForm(form => ({
+      _csrf: `${window.session._csrf}`,
+      ...form,
+      [e.target.name]:
+        e.target.type === 'checkbox' ? e.target.checked : e.target.value,
+    }))
+  }
+  const { error } = schema.validate({ ...form })
+  const isValid = error === undefined
```
kasbah, line 16: `== null`

frontend/pages/_app.js
```diff
@@ -2,19 +2,7 @@ import App from 'next/app'
...
-import 'semantic-ui-css/components/container.css'
-import 'semantic-ui-css/components/divider.css'
-import 'semantic-ui-css/components/grid.css'
-import 'semantic-ui-css/components/icon.css'
-import 'semantic-ui-css/components/image.css'
-import 'semantic-ui-css/components/input.css'
-import 'semantic-ui-css/components/menu.css'
-import 'semantic-ui-css/components/message.css'
-import 'semantic-ui-css/components/popup.css'
+import 'semantic-ui-css/semantic.min.css'
```
kasbah, line 5: I'd like to keep importing these individually. I know it's a pain but it does reduce the size of the CSS.

frontend/components/NavBar.js
```diff
@@ -0,0 +1,239 @@
+import React, { useContext } from 'react'
+import { Button, Icon, Image, Menu, Popup } from 'semantic-ui-react'
+import { useRouter } from 'next/router'
+
+import { AuthContext } from '../contexts/AuthContext'
+import styles from './TitleBar.module.scss'
```
kasbah, line 6: Should rename it here too.

frontend/components/TitleBar.js
```diff
@@ -1,115 +1,142 @@
 import React from 'react'
```
kasbah, line 1: I guess this file should be deleted?

frontend/cypress/integration/authRedirects.spec.js
```diff
@@ -0,0 +1,28 @@
...
+  const password = '123456'
+
+  before(() => {
+    cy.visit('/')
+    cy.createUser(username, email, password)
+  })
+  it("Redirects unauthenticated users to '/login' when accessing reqSignIn page", () => {
+    // `/project/new` is marked as `reqSignIn`.
+    cy.visit('/projects/new')
+    cy.url().should('eq', 'http://kitspace.test:3000/login')
```
kasbah, line 15: Can we add a test that it re-directs back to the requested page (`/projects/new`) after login?

frontend/cypress/integration/authRedirects.spec.js
```diff
@@ -0,0 +1,28 @@
...
+  })
+
+  it('Redirects authenticated users to homepage when accessing reqSignOut page', () => {
```
kasbah, line 18: I think we should re-direct to /login when signing out. The main reason someone signs out manually is because they want to sign in with a different account.

frontend/cypress/integration/navbar.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+    // sign the user in.
+    cy.stubSignInReq(true, { LoggedInSuccessfully: true })
+    cy.signIn(username, password)
+
+    // After a successful login the user is redirect to the homepage.
+    cy.url().should('eq', 'http://kitspace.test:3000/')
+  })
+
+  it('Redirect to correct page if there is a redirect query', () => {
```
kasbah, line 62: "it redirect*s*"

frontend/cypress/integration/navbar.spec.js
```diff
@@ -0,0 +1,79 @@
...
+  before(() => {
+    cy.visit('/')
+    cy.createUser(username, email, password)
+  })
+
+  beforeEach(() => {
+    cy.signOut()
+  })
+
+  it('unauthenticated user', () => {
```
kasbah, line 17: Would be good to write all these test names as grammatically correct sentences beginning with "it: "it does something..."

frontend/cypress/integration/navbar.spec.js
```diff
@@ -0,0 +1,79 @@
...
+  it('authenticated user', () => {
+    // sign the user in.
+    cy.visit('/login')
+    cy.stubSignInReq(true, { LoggedInSuccessfully: true })
+    cy.signIn(username, password)
+
+    // Clicking `Add Project` redirects to new project page.
+    cy.get('#add_project').click()
+    cy.url().should('eq', 'http://kitspace.test:3000/projects/new')
+  })
```
kasbah, line 35: Why is an authenticated user signing back in? I don't think we should allow that behaviour, rather say to the user: you are already signed in.

frontend/cypress/support/commands.js
AbdulrhmnGhanem: It's enforced by the Cypress directory structure. The support root file can be configured to be any other file, though I prefer leaving it as is.

frontend/pages/_app.js
AbdulrhmnGhanem: It's on my Todo though I have to figure out what to import actually :sweat_smile:

frontend/cypress/integration/navbar.spec.js
AbdulrhmnGhanem: To test the behavior for authenticated users, it needs an authenticated user. So it creates a random user and sign him in and proceed to check the actual logic. It's more of a setup step, not the test itself.

frontend/cypress/integration/authRedirects.spec.js
AbdulrhmnGhanem: Easy fix!

frontend/components/Page.js
AbdulrhmnGhanem: See 5cbd357, 7d59eb1, cd36ff5, 50f395c. I solved this in `import-projects` as it was part of another problem.

frontend/cypress/support/commands.js
kasbah: I see, that's fine then. What does it mean though just out of interest?

README.md
```diff
@@ -34,27 +34,26 @@ git submodule update --init
...
 
-7. Build and bring up the frontend 
-
-```
-docker-compose up frontend
-```
+3. Source the `.env` file with `set -a && source ../.env && set +a`
 
-8. Bring up the rest of the containers (in the future this command is all you need to do to bring up everything, including gitea and the frontend)
+3. Start the testing client with the `npx cypress open` command.
```
kasbah, line 55: Looks like you have steps 3 now.

frontend/cypress/support/commands.js
AbdulrhmnGhanem: It contains utility functions though instead of accessing them directly from the files having them in the `support` makes them accessible as properties of the `Cypress` object.

README.md
AbdulrhmnGhanem: What?

kasbah: > 3. Source the `.env` file with ...

> 3. Start the testing client with the ...

kasbah:
Noticed an issue today. You can create an account, but you stay on the page. Then if you change the password you can get some contradictory messages. 

image

PR #19, opened 2020-08-24
Optimize Docker build

AbdulrhmnGhanem:
- Initially, it was 1.16GB
- Now it is 334MB
- Further optimization should be done to the production version.
Even if it doesn't make a difference to the performance yet my hard drive is grateful.


kasbah:
Looks good to me. Let me just test the updated gitea first and then I'll test this too and merge it in.

PR #26, opened 2020-09-13
[WIP] Rewrite Navbar

AbdulrhmnGhanem:
WIP for #17 
* Now the navbar is user aware, you can try it by toggling the `AuthProvider` `state` from React dev tools.
* I think the CSS style needs some improvements.


frontend/components/NavBar.js
```diff
@@ -195,17 +195,21 @@ function LoginButton() {
...
+
+    xhr.open('POST', endpoint, true)
+    xhr.setRequestHeader('Content-Type', 'application/x-www-form-urlencoded')
+    xhr.onreadystatechange = () => {
+      if (xhr.readyState === XMLHttpRequest.DONE && xhr.status === 200) {
+        router.push('/')
+      }
+    }
+
+    xhr.send(`_csrf=${window.session._csrf}`)
```
kasbah, line 212: Why are you switching to XHR!?

AbdulrhmnGhanem: From the commit message:

- The sign out mechanism expects a request done with url-encoded-form, also using xhr not fetch.
- The headers get set correctly though `i_like_gitea`, `_csrf`, and `lang` for some reason get set twice in the response.
- One of them is from the sign out endpoint which clear them(the headers).
- Haven't identified the second source yet. It sets a value to them which makes the sign out fails.

I have been debugging it for a while now not sure if `fetch` is one of the reason, will try to get it to work with xhr then it will be easy to migrate to fetch.

AbdulrhmnGhanem: Calling the original logout endpoint solved the problem. I Will see if it can be done with fetch.

PR #30, opened 2020-10-30
Import projects

AbdulrhmnGhanem:
Importing project either by syncing or by files upload


AbdulrhmnGhanem:
Just from a quick glance:
1. Create a repo, which is part of the official API. In the future probably it needs to do more than that, say a user is trying to update an existing project, creating a new repo for each upload isn't the optimal solution.
2. It seems there's an endpoint for [creating](https://github.com/kitspace/gitea/blob/bfb404274f4a48e7c058ffabcb9f103d161f8594/routers/routes/routes.go#L795)/[uploading](https://github.com/kitspace/gitea/blob/bfb404274f4a48e7c058ffabcb9f103d161f8594/routers/routes/routes.go#L800) files. Not sure how it works yet!

kasbah:
Can you explain what you mean by `1.`? Is there a problem with something you are trying to do, or something else? I don't understand...

Regarding 2., there are official API docs here by the way: https://try.gitea.io/api/swagger

AbdulrhmnGhanem:
> Regarding 2., there are official API docs here by the way: https://try.gitea.io/api/swagger (should also be available on http://gitea.kitspace.test:3000/api/swagger)

It isn't part of the API, or at least I couldn't find it.
The file upload page in Gitea makes a request to `/{username}/{repo}/_upload/{branch}/{directory}`. It sends a `x-www-form-urlencoded` request with files. Also, it performs some kind of hashing on the files before upload.

> Can you explain what you mean by `1.`? Is there a problem with something you are trying to do, or something else? I don't understand...

Not now, I am assuming a scenario where a user has already uploaded a project and he tries updating it. Should we update the repo or create a new one?

kasbah:
Definitely update in that scenario, yeah. But updating shouldn't be possible through the `/projects/new` endpoint. 

Seems there is an endpoint to create a single file: https://try.gitea.io/api/swagger#/repository/repoCreateFile, maybe we should create an endpoint to create multiple files via the API and upstream it.

AbdulrhmnGhanem:
The cors error appeared once again!

AbdulrhmnGhanem:
NVM solved it! The same rule which applies to `'Access-Control-Allow-Origin'` also applies to `'Access-Control-Allow-Credentials'`. 

**The rule is the header gets only set for 2xx responses**

AbdulrhmnGhanem:
Now it throws `bad request`/ `unauthorized` from gitea while actually being authorized!

AbdulrhmnGhanem:
Have you tried [repoCreateFile](https://try.gitea.io/api/swagger#/repository/repoCreateFile) before? When I call it from the dashboard it throws a 404 error! I'm not sure what is the `path`! Is it both the branch name and file name? or just filename? I tried both neither worked. While calling it from the browser returns an internal server error with no message. I can't get what is missing.

AbdulrhmnGhanem:
I have considered creating a gitea endpoint for accepting multiple files, though I won't be a simple change; as for now, the URL itself to which the request gets sent contains the file path.

AbdulrhmnGhanem:
```
2020/11/08 00:31:32 ...pofiles/temp_repo.go:254:Push() [I] Unable to push back to repo from temporary repo due to rejection: testuser/testRepo12 (/data/gitea/tmp/local-repo/upload.git003504194)
	Stdout: 
	Stderr: remote: error: cannot lock ref 'refs/heads/master': is at bf20739bf2b47d329115744891b5b8f3cbe09f99 but expected 7db476e6b0518f6d0b212ea6ae9f049ed232b59e        
	To /data/git/repositories/testuser/testrepo12.git
	 ! [remote rejected] cc0645767540dee3d2d9aeaad13b34abfbe4ca8e -> master (failed to update ref)
	error: failed to push some refs to '/data/git/repositories/testuser/testrepo12.git'
	
	Error: exit status 1
2020/11/08 00:31:32 .../repofiles/update.go:449:CreateOrUpdateRepoFile() [E] *log.ColoredValue PushRejected Error: exit status 1: remote: error: cannot lock ref 'refs/heads/master': is at bf20739bf2b47d329115744891b5b8f3cbe09f99 but expected 7db476e6b0518f6d0b212ea6ae9f049ed232b59e        
	To /data/git/repositories/testuser/testrepo12.git
	 ! [remote rejected] cc0645767540dee3d2d9aeaad13b34abfbe4ca8e -> master (failed to update ref)
	error: failed to push some refs to '/data/git/repositories/testuser/testrepo12.git'
```
I believe it's a race condition! I will try to send the requests sequentially and see what will happen.

AbdulrhmnGhanem:
I didn't like the modal, so I removed it though I hated it :sweat_smile:, I reverted the change back. Anyways you can check out both commits and see which one you like.

AbdulrhmnGhanem:
Does a gitea have an endpoint for fetching files from a repo?

AbdulrhmnGhanem:
Found it `repository/repoGetContentsList`!

AbdulrhmnGhanem:
Should we have a `my-projects` page? at least for now as there's no way to get the projects so it can be modified except for going for the URL directly.

AbdulrhmnGhanem:
For some reason which I couldn't figure the update page only behaves as expected if you drop files into it.

AbdulrhmnGhanem:
> reason

It was a CSS bug :sweat_smile: 


- If a gitea repo exists `{user}/{repo}` then if you visit `projects/update/{user}/{repo}` it should display the project details in the update form. If you load files it will add them to the pre-existing files in the preview component. 
- Though the update button isn't functional yet.
- Dropping files at `/projects/new/` redirect to `projects/update/{user}/{repo}/new` though it will display 404 as this page works with existing repos only. You can remove this part and it will display the dropped files in the update page.

kasbah:
Did a quick test today, encountered an error (`files` can be null) I guess that bit isn't done yet. Looks like you are making good progress. Let me know if you think something is ready to test. By the way we should be able to upload a whole folder when selecting files.

AbdulrhmnGhanem:
> Did a quick test today, encountered an error (`files` can be null) I guess that bit isn't done yet.

Please, describe the steps to reproduce the error.

kasbah:
Happens just when I upload a file on the right, or drop a folder in there.

kasbah:
```
Unhandled Runtime Error

TypeError: files is null
Source

contexts/UploadContext.js (38:12) @ _callee2$

  36 |   const repo = projectNameFromPath(asPath)
  37 |   const files = await getRepoFiles(repo, csrf)
> 38 |   const filesDetails = files.map(({ name, size }) => ({ name, size }))
     |        ^
  39 |   setRepoFiles(filesDetails)
  40 | }
  41 | if (isUpdateRoute && !fetchedRemote) {
```

frontend/pages/projects/mine.js
```diff
@@ -1,27 +1,95 @@
...
+import styles from './mine.module.scss'
+
+const DeleteModal = ({ projectName }) => {
+  const { reload } = useRouter()
+  const { csrf } = useContext(AuthContext)
+  console.log(projectName)
+  return (
+    <Modal
+      trigger={<Button content="Delete" color="red" />}
+      heade="Heads up!"
```
kasbah, line 18: Missing an `r` in `header`.

kasbah:
Ran into an error when pressing "log out" from the /projects/new page. 


```
Unhandled Runtime Error

TypeError: user is null
Source

pages/projects/new.js (41:14) @ Sync

  39 |   const [remoteRepo, setRemoteRepo] = useState('')
  40 | 
> 41 |   const uid = user.id
     |              ^
  42 |   const username = user.login
  43 | 
  44 |   const handleClick = async () => {
```

frontend/utils/giteaApi.js
```diff
@@ -0,0 +1,229 @@
...
+  const reqBody = {
+    author: {
+      email: user.email,
+      name: user.login,
+    },
+    branch: 'master',
+    committer: {
+      email: 'admins@kitspace.org',
+      name: 'Kitspace',
+    },
```
kasbah, line 195: I think this should be the same as author.

frontend/utils/giteaApi.js
```diff
@@ -0,0 +1,229 @@
...
+    // content must be Base64 encoded
+    content: btoa(content),
+    message: `Automated commit on behalf of ${user.login} (${user.email})`,
```
kasbah, line 198: I think we can do better, that message is too long. I would suggest just `Upload {path}` when uploading a single file and `Upload multiple files` when we have that.

frontend/utils/giteaApi.js
```diff
@@ -0,0 +1,229 @@
...
+}
+
+/**
+ * Get list of files in a gitea repo.
+ * @param repo {string}
+ * @param csrf {string}
+ * @returns {Promise<Array|null>}
+ */
+export const getRepoFiles = async (repo, csrf) => {
+  const endpoint = `${giteaApiUrl}/repos/${repo}/contents?ref=master&_csrf=${csrf}`
```
kasbah, line 122: Just leaving a note here: we can't assume the default branch will be `master`.

frontend/utils/giteaApi.js
AbdulrhmnGhanem: I think this is semantically more [correct](https://stackoverflow.com/a/23108169/12113647).

frontend/utils/giteaApi.js
AbdulrhmnGhanem: This repo is created when a user uploads files through drag and drop **not** through `sync`. So I guess we are free to upload to whichever branch we want. We shouldn't guess the branch from which the assets will be generated but this will be a separate feature.

frontend/utils/giteaApi.js
kasbah: No, the way I see it. People are using Kitspace author and commit their code. You won't see Github as "commiter" either when you use the web interface.

frontend/utils/giteaApi.js
kasbah: So `getRepoFiles` won't work on repos that are synced?

frontend/utils/giteaApi.js
kasbah: Maybe even just `Add <filename>`, that's what Gitea seems to do.

frontend/pages/projects/new.js
```diff
@@ -45,14 +47,26 @@ const Upload = ({ user, csrf }) => {
...
+      <DropZone onDrop={onDrop} />
+      <Modal
+        closeIcon
+        header="Heads up!"
+        content="You have an existing project with the same name, do you want to overwrite it?"
+        open={open}
+        onClose={() => setOpen(false)}
+      />
+    </>
+  )
```
kasbah, line 69: I think we shouldn't offer to overwrite a project when uploading a single file, rather the options should be

1. Update an existing project
2. Create a new project with a different name

frontend/utils/giteaApi.js
AbdulrhmnGhanem: Oh, sorry I thought the commit was on a createRepo. You're right.

AbdulrhmnGhanem:
Tests for the file upload seems a bit challenging! Please, take a look whenever you can and let me know if you have any suggestions.

frontend/utils/giteaApi.js
AbdulrhmnGhanem: - Added optional param for the branch in 9513187.
- Though haven't updated the code that depends on `getRepoFiles` to fetch from the main branch yet.
- We might take the same approach as GitHub pages which builds the website from the main branch and ignores the rest.

p.s: I tried `HEAD`, it didn't work.

kasbah:
What's the main challenge? `cypress-file-upload` doesn't work well then?

AbdulrhmnGhanem:
> What's the main challenge? `cypress-file-upload` doesn't work well then?

Yeah, on drop either a redirect should happen(no conflicts) or a modal showed appear(conflict). Nither worked as expected.

kasbah:
The new warning is a bit too fast. I just type `h` and immediately get this. I think it should only show up when you press the button. (It also shouldn't move the input so I think it should be at the bottom.)

image

Also there's a problem with the style on a small window width.

image

kasbah:
Folder upload doesn't seem to be working correctly yet. I drag a folder and it only creates a repo with a default LICENSE and README.

kasbah:
On Gitea it looks like this

image

But on Kitspace like this

image

AbdulrhmnGhanem:
> Folder upload doesn't seem to be working correctly yet. I drag a folder and it only creates a repo with a default LICENSE and README.

It's also broken for files, not just folders I broke it few commits away will fix it once I get the multi-file upload done. It's pretty simple I believe.

frontend/src/utils/giteaInternalApi.js
```diff
@@ -1,8 +1,14 @@
+/*
+* Committing multiple files to gitea is a two steps process:
+*  1. Upload the files to gitea which it turn returns a UUID for that file
+*  2. UUID returned from the previous step is used to make the actual commit request.
+*/
+
 /**
- *
- * @param repo
+ * Upload a file to gitea server, just upload it doesn't commit the files
```
kasbah, line 8: Huh? What's your plan here? I think you might over-complicating this? I think we should upload and commit multiple files in one request.

AbdulrhmnGhanem:
It's the same approach used by gitea GUI

AbdulrhmnGhanem:
This enables concurrent upload of the files.

frontend/src/pages/projects/update/[user]/[projectName].js
```diff
@@ -0,0 +1,127 @@
...
+  const fullname = `${user}/${projectName}`
+
+  useEffect(() => {
+    const getProject = async () => {
+      const project = await getRepo(fullname)
+      setProject(project)
+    }
+    setPersistenceScope(projectName)
+    getProject().then()
+  }, [])
```
Comment (kasbah) [inline frontend/src/pages/projects/update/[user]/[projectName].js:28]: Everywhere were you have an empty `.then()` something strange is going on.

1. An empty `.then()` is a no-op, not sure why you have added these at all
2. Most of the time your use-effects can be simplified and made clearer by avoiding async/await, e.g this one could be:

```js
  useEffect(() => {
    setPersistenceScope(projectName)
    getRepo(fullname).then(setProject)
  }, [])
```

frontend/src/pages/projects/update/[user]/[projectName].js
```diff
@@ -7,7 +7,7 @@ import FilesPreview from '@components/FilesPreview'
 import useForm from '@hooks/useForm'
 import { ProjectUpdateForm } from '@models/ProjectUpdateForm'
 import { UploadContext } from '@contexts/UploadContext'
-import { getRepo, updateRepo } from '@utils/giteaApi'
+import { getRepo as getProject, updateRepo } from '@utils/giteaApi'
```
Comment (kasbah) [inline frontend/src/pages/projects/update/[user]/[projectName].js:10]: image

Comment (AbdulrhmnGhanem) [reply frontend/src/pages/projects/update/[user]/[projectName].js:10]: :rofl: :rofl: :rofl:
I think it makes more sense; in the pages, the repo is always referenced as a project, though In the API client it's explicit what endpoint we are hitting

Comment (kasbah) [reply frontend/src/pages/projects/update/[user]/[projectName].js:10]: I think it's and extra level of renaming that makes it less clear. Use the API you are writing (the Gitea client). If you don't like it, write a better API.

AbdulrhmnGhanem:
Are you sure you got both folder and file selector to work at the same time? I can't find anything that does so in vanilla HTML. It's stated [here](https://stackoverflow.com/questions/42633306/how-to-allow-the-user-to-pick-any-file-or-directory-in-an-input-type-file-ta) it's not possible. Also, some services such as Google Drive have separate pickers for folders and files. Should I add another button or a selector to enable different picking modes?

kasbah:
No, I only got directory upload working. Separate buttons it is then. I can't believe web-tech sometimes.

frontend/src/pages/projects/update/[user]/[projectName].js
Comment (kasbah) [reply frontend/src/pages/projects/update/[user]/[projectName].js]: You still need to do a search and delete any remaining empty `.then()`.

PR #31, opened 2020-12-11
Attempt to use drone ci to process assets

kasbah:
Had this idea that we could maybe use [drone](https://github.com/drone/drone/) to process the files we need. So far I mostly got stuck on on getting Oauth hooked up and resolving domains. It may not be possible to run drone on localhost properly.

### DIFF
--- a/docker-compose.production.yml
+++ b/docker-compose.production.yml
@@ -12,7 +12,7 @@ services:
   nginx:
     environment:
       - "RUN_CERTBOT=true"
-      - CERTBOT_DOMAINS=${KITSPACE_DOMAIN},gitea.${KITSPACE_DOMAIN}
+      - CERTBOT_DOMAINS=${KITSPACE_DOMAIN},gitea.${KITSPACE_DOMAIN},drone.${KITSPACE_DOMAIN}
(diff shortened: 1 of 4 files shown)


kasbah:
I got this running on https://drone.staging.kitspace.org 

- We could maybe write a [conversion extension](https://docs.drone.io/extensions/conversion/) to take kitspace.yaml files and process them into a .drone.yml.
- There's an API to get the build status: https://docs.drone.io/api/user/
- Other the other side drone seems to have several plugins to upload artifacts to S3:
    - http://plugins.drone.io/drone-plugins/drone-s3/
    - http://plugins.drone.io/drone-plugins/drone-s3-sync/

Just the OAuth requirements and DNS really don't seem to want to work on localhost, which is annoying.

kasbah:
I evaluated this and also [Agola](https://github.com/agola-io/agola) and [Councourse](https://github.com/concourse/concourse). I think we complicate things by using a service like this that isn't quite designed for our purposes. 

I am going to make a NodeJS server called `process` instead which will process the assets and serve them up or, down the line, put them into s3.

PR #36, opened 2020-12-20
Add a gitea dev docker container config

### LINKED ISSUE #23: Figure out how to effeciently re-compile gitea in docker for development
It takes ages to re-compile at the moment.

kasbah:
closes #23 

With this, `docker-compose up gitea` will run `make backend` before starting gitea, so our changes to gitea will be compiled in an effecient way.

### DIFF
--- a/Dockerfile.gitea.dev
+++ b/Dockerfile.gitea.dev
@@ -1,0 +1,63 @@
+
+###################################
+#Build stage
+FROM golang:1.15-alpine3.12
+
+ARG GOPROXY
+ENV GOPROXY ${GOPROXY:-direct}
+
+ARG GITEA_VERSION
+ARG TAGS="sqlite sqlite_unlock_notify"
+ENV TAGS "bindata timetzdata $TAGS"
+ARG CGO_EXTRA_CFLAGS
+
+#Build deps
+RUN apk --no-cache add build-base git nodejs npm
+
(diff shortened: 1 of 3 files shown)


AbdulrhmnGhanem:
I don't understand why will this work? AFICR, `COPY` is used to get the source code into docker which will invalidate all the subsequent cached layers when a new `build` command is issued.

kasbah:
The way it works is:

1. You run `docker-compose build gitea` if you want to run the whole compilation from scratch. 
2. You Ctrl-C and run `docker-compose up gitea` if you just want to re-compile the gitea "backend" quickly (and Make knows which files actually need re-compiling)

I may have done something strange, as I only half know what I am doing here, but it does seem to work. The initial build COPY and build is mounted as a volume and that seems to work out.

AbdulrhmnGhanem:
This was the first thing that came to my mind but I immediately discarded it because of the `COPY` command. The screenshot below from Docker docs [/best-practices/use-multi-stage-builds](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#use-multi-stage-builds)
image

kasbah:
It's probably a bit strange to copy everything into the container and then mount that as a volume into the same place but I just tried to keep it as close as possible to the original Gitea dockerfile.

PR #45, opened 2020-12-24
Rewrite tests in Playwright

AbdulrhmnGhanem:
Cypress turned out a bit crippling, will try playwright and see if it will eliminate the pain points from cypress


kasbah:
This PR is against a dead branch.

PR #53, opened 2020-12-31
Publish docker images to ghcr and make use of them

kasbah:
Making this pull against master, following on from #47

PR #54, opened 2021-01-01
Add production ready Docker files

AbdulrhmnGhanem:
- Add separate docker files configured to work in production for
  - frontend
  - processor

- Update the compose files accordingly.


AbdulrhmnGhanem:
Now there are two docker-compose files with similar behavior `docker-compose.production.yml` and `docker-compose.deploy.yml`. Shall I delete the `docker-compose.production.yml`? If it has a role I don't know then it probably needs a different name.

kasbah:
When it works, then yes we should replace the production one. Why are we doing this on two branches? Let's keep working on it in #53.

Issue #37, opened 2020-12-22
Need to be able to use Gitea API server-side in frontend


kasbah:
We need to be able to use the Gitea API in `getInitialProps` so we can server-render pages.

kasbah:
I think we should [swr](https://swr.vercel.app/) for caching and feed it in from `getInitialProps` like they suggest: https://swr.vercel.app/docs/with-nextjs

Issue #56, opened 2021-01-02
Nginx timeout when syncing big repos


AbdulrhmnGhanem:
Syncing a big repository e.g., [ulx3s](https://github.com/emard/ulx3s) takes a while which makes Nginx returns `504`. Currently, this is considered a failure, and an error message gets displayed as if syncing failed.

kasbah:
I think #58 is a bit of a workaround. I don't think it's reasonable of Gitea to wait to the import to complete before returning the response. It should probably be fixed upstream. I think we should also re-direct to a page that says "Importing..." , ignore the initial http response and then poll for completion. (I think that's how Gitea does it actually?)

AbdulrhmnGhanem:
I couldn't find a way to migrate from the GUI. Please, upload a screenshot.

kasbah:
Here, though it's not that important what Gitea itself does. It doesn't use the API so the response seems to be a 302 re-direct. 

![image](https://user-images.githubusercontent.com/206854/103463271-31d26c00-4d23-11eb-8162-782abd69ffa0.png)

AbdulrhmnGhanem:
> It doesn't use the API so the response seems to be a 302 re-direct.

Yes, It uses an internal endpoint `/{repo}/status` which returns `{errors: "", status: 1}` while the repo is still syncing  and ` {errors: "", status: 0}` when it's done. I looked into the source and it doesn't provide progress status.


> It doesn't use the API so the response seems to be a 302 re-direct.

This shold do the trick
```js
window.addEventListener("beforeunload", function() { debugger; }, false)
```

AbdulrhmnGhanem:
This problem is less severe in the staging server; the network is better than my crappy one :sweat_smile:

kasbah:
> This should do the trick
>> window.addEventListener("beforeunload", function() { debugger; }, false)

Huh, TIL about `debugger`, I used the network debug tab though which works fine too. 

> This problem is less severe in the staging server; the network is better than my crappy one 

Sure, we just have to find bigger git repos :)

AbdulrhmnGhanem:
> Yes, It uses an internal endpoint `/{repo}/status` which returns `{errors: "", status: 1}` while the repo is still syncing and ` {errors: "", status: 0}` when it's done. I looked into the source and it doesn't provide progress status.

This won't work out of the box for repositories migrated by the API:
[...]

PR #57, opened 2021-01-02
Static generation and SWR support

AbdulrhmnGhanem:
Statically generate the `index` and repo pages at build time.
Currently, there's an issue with the DNS; the first commit demonstrate a try to add static generation to the `index`. It fails with this error message
image


frontend/src/pages/index.js
```diff
@@ -40,6 +42,30 @@ const ProjectCard = ({ name, full_name, description, owner }) => {
   )
 }
 
+export const getStaticProps = async () => {
```
AbdulrhmnGhanem, line 45: I know there's a function in the gitea client doing the same thing though I preferred to make the commit in a single file

frontend/src/pages/index.js
```diff
@@ -40,6 +42,30 @@ const ProjectCard = ({ name, full_name, description, owner }) => {
...
+    }
+  } else {
+    return {
+      props: {
+        repos: [],
+      },
+    }
+  }
+}
+
```
AbdulrhmnGhanem, line 68: I know there's a function in the gitea client doing the same thing though I preferred to make the commit in a single file

kasbah:
If you update your /etc/hosts, I think it will sort out the DNS issue. 

```
172.17.0.1    kitspace.test
172.17.0.1    gitea.kitspace.test
172.17.0.1    processor.kitspace.test
```

AbdulrhmnGhanem:
> If you update your /etc/hosts, I think it will sort out the DNS issue.

I did so, does it need a system reboot?

kasbah:
Hm, don't think it does.

kasbah:
I pushed a fix, it adds `network_mode: host` to the docker-compose.

AbdulrhmnGhanem:
The frontend crashes with this error now
```log
frontend_1   |       throw er; // Unhandled 'error' event
frontend_1   |       ^
frontend_1   | 
frontend_1   | Error: listen EADDRINUSE: address already in use :::3000
frontend_1   |     at Server.setupListenHandle [as _listen2] (net.js:1279:14)
frontend_1   |     at listenInCluster (net.js:1327:12)
frontend_1   |     at Server.listen (net.js:1414:7)
frontend_1   |     at Function.listen (/app/node_modules/express/lib/application.js:618:24)
frontend_1   |     at app.prepare.then (/app/server.js:36:10)
frontend_1   | Emitted 'error' event at:
frontend_1   |     at emitErrorNT (net.js:1306:8)
frontend_1   |     at process._tickCallback (internal/process/next_tick.js:63:19)
```

Running `ss -plunt | grep :3000` shows no processes on this port

AbdulrhmnGhanem:
I am not sure what is going on, it might be the internal ports on the docker network colliding with port on the host. I switched to user [defined networks](https://docs.docker.com/compose/networking/#specify-custom-networks) still building :crossed_fingers:

kasbah:
Hmm, weird I thought the `network_mode` thing worked for me at first, now it's the same as with yours. 

I think the solution is to mount /etc/hosts into the frontend. I've reverted the other changes and pushed that. With that I think the DNS is working, it now fails on a different issue after fetching the data successfully I believe. 

```
Error: Error serializing `.repos` returned from `getStaticProps` in "/".
Reason: `undefined` cannot be serialized as JSON. Please use `null` or omit this value all together.
```

AbdulrhmnGhanem:
closes #37

docker-compose.yml
```diff
@@ -40,7 +42,7 @@ services:
       - '2222:22'
     depends_on:
       - postgres
-    command: /bin/bash --verbose -c "cd /go/src/code.gitea.io/gitea/ && make backend && cp gitea /app/gitea/ && /usr/bin/entrypoint"
+#    command: /bin/bash --verbose -c "cd /go/src/code.gitea.io/gitea/ && make backend && cp gitea /app/gitea/ && /usr/bin/entrypoint"
 
   postgres:
     image: postgres:9.6
```
kasbah, line 48: Please remove these docker-compose changes from this branch, we'll figure this out separately.

frontend/src/pages/index.js
```diff
@@ -1,56 +1,28 @@
...
-  const [projects, setProjects] = useState([])
 
-  useEffect(() => {
-    if (csrf) {
-      getRepos(csrf).then(setProjects)
-    }
-  }, [csrf])
+  const { data: projects } = useSWR('/', getAllRepos, { initialData: repos })
 
+  // TODO: handle the failure case i.e., projects is empty.
```
kasbah, line 25: You shouldn't have TODOs like this if you say it's ready to merge.

frontend/src/utils/giteaApi.js
```diff
@@ -170,21 +170,19 @@ export const getRepo = async fullname => {
...
  * @param order{string}
- * @param q{string}: search query, leave undefined to return all repos
- * @returns {Promise<any|*[]>}
+ * @param q{string=}: search query, leave undefined to return all repos
+ * @returns {Promise<[Object]>}
  */
-export const searchRepos = async (csrf, sort = 'updated', order = 'desc', q) => {
-  const endpoint = `${giteaApiUrl}/repos/search?_csrf=${csrf}`
+export const searchRepos = async (sort = 'updated', order = 'desc', q) => {
+  const endpoint = `${giteaApiUrl}/repos/search`
```
kasbah, line 185: What's happening with the csrf? Why isn't it needed?

frontend/src/utils/index.js
```diff
@@ -19,3 +20,22 @@ export const slugifiedNameFromFiles = files => {
...
+}
+
+/**
+ * Get the project name from the `path` object in `next.router`.
+ * @param path
+ * @returns {*}
+ */
+export const projectNameFromPath = path => {
+  return path.split('/').slice(3).join('/')
+}
```
kasbah, line 41: Would be good to have an example in each of these, something like this (if it's correct);
```js
/*

e.g. /user/project/x/y -> /user/project 

*/
```

frontend/src/pages/index.js
AbdulrhmnGhanem: I wasn't sure what to do in this case: it won't fail or raise any error, the `index` page won't have any projects just like this. Should the build deliberately fail in production in this case? display another component saying "Nothing here"? Or just leave it as is?

frontend/src/utils/giteaApi.js
AbdulrhmnGhanem: See [CSRF with GET requests](https://security.stackexchange.com/questions/115794/should-i-use-csrf-protection-for-get-requests). Luckily, Gitea got this part right and it follows the right semantics of `GET` requests; it doesn't perform any action just get the publicly available data. All these endpoints don't even require authentication to get a response from them.

frontend/src/pages/index.js
kasbah: Ah, I see. I thought you meant it crashes with no projects. Empty page is fine for now, given this shouldn't really happen. Once we add search it will be different again anyway.

frontend/src/pages/index.js
```diff
@@ -1,56 +1,28 @@
...
+  const { user } = useContext(AuthContext)
   const username = user?.login || 'unknown user'
-  const [projects, setProjects] = useState([])
 
-  useEffect(() => {
-    if (csrf) {
-      getRepos(csrf).then(setProjects)
-    }
-  }, [csrf])
+  const { data: projects } = useSWR('/', getAllRepos, { initialData: repos })
```
kasbah, line 23: Here

frontend/src/pages/projects/[user].js
```diff
@@ -0,0 +1,77 @@
...
+      username: params.user,
+    },
+  }
+}
+
+const User = ({ userRepos, username }) => {
+  const { user } = useContext(AuthContext)
+  const { replace } = useRouter()
+
+  const { data: projects } = useSWR(username, getUserRepos, {
```
Comment (kasbah) [inline frontend/src/pages/projects/[user].js:36]: Here

frontend/src/pages/projects/mine.js
```diff
@@ -1,48 +1,38 @@
...
-          onClick: async () => {
-            await deleteRepo(projectName, csrf)
-            await reload()
-          },
-        },
-      ]}
-    />
-  )
+const useAuthUserProjects = user => {
+  const { data, error } = useSWR(user, getUserRepos)
```
kasbah, line 15: Here

frontend/src/pages/projects/update/[user]/[projectName].js
```diff
@@ -144,4 +163,14 @@ const UpdateForm = ({ isNew, withDropZone, owner, name, description }) => {
   )
 }
 
+const useUpdateFrom = projectName => {
+  const { data, error } = useSWR(projectName, getRepo)
```
Comment (kasbah) [inline frontend/src/pages/projects/update/[user]/[projectName].js:167]: Here

frontend/src/components/ProjectCard.js
```diff
@@ -0,0 +1,47 @@
...
+import styles from './ProjectCard.module.scss'
+
+const processorUrl = process.env.KITSPACE_PROCESSOR_URL
+
+const fetcher = (...args) => fetch(...args).then(r => r.json())
+
+const useThumbnail = full_name => {
+  const img = `/${full_name}/HEAD/images/top.png`
+  const statusUrl = processorUrl + '/status/' + img
+  const { data, error } = useSWR(statusUrl, fetcher, { refreshInterval: 1000 })
```
kasbah, line 15: This was how I did it. I think this makes more sense.

frontend/src/contexts/UploadContext.js
```diff
@@ -105,9 +98,28 @@ export default function UploadContextProvider(props) {
...
+/**
+ * Fetch the files in the gitea repo associated with project being updated
+ * @param repo{string}
+ * @param shouldFetch{boolean}
+ * @returns {{isLoading: boolean, isError: boolean, files: any | Array | null}}
+ */
+const useFetchRemoteRepoFiles = (repo, shouldFetch) => {
+  const fetcher = () => getDefaultBranchFiles(repo)
+
+  const { data, error } = useSWR(shouldFetch ? `update/${repo}` : null, fetcher)
```
kasbah, line 118: This is a really weird one.

kasbah (changes requested):
The way you are using keys in SWR doesn't make sense to me. I think we will trip ourselves up with the inconsistency of these. 

It's a global cache so the keys should reflect the exact data being cached. That will enable us to pull down that data anywhere in the app. 

I suggest using the full API URL with encoded query params as the key. This may mean re-thinking the structure of the Gitea client.

frontend/src/contexts/UploadContext.js
AbdulrhmnGhanem: It's exactly as the [docs](https://swr.vercel.app/docs/conditional-fetching) say how it should be done :sweat_smile:

kasbah: Oh, haha, hadn't seen that. Could still use a comment I think. Also, what's going on here, why do we call `useFetchRemoteRepoFiles` when we don't actually want the files?

The rest of what I said about the keys still stands anyway.

AbdulrhmnGhanem:
This will mean bypass the Gitea client altogether for these requests, I think this will make the pages really involved in the interface with Gitea. For example, the `getAllRepos` function technically is a hack. I believe exposing this level of detail will trip us even worse.  You are right about the inconsistencies but I don't think this is the right solution.

frontend/src/contexts/UploadContext.js
AbdulrhmnGhanem: The value of `shouldFetch` is true if the current page is under the `projects/update` route. But `UploadContext` wraps the `Page` component so It will make unnecessary requests, it won't do any harm to make the request apart from some `404`s responses.

kasbah:
What if we refactor the client to export SWR React hooks that we import everywhere else? So instead of `getAllRepos` we export `useFetchAllRepos` and use sensible keys within that.

AbdulrhmnGhanem:
OK, this seems better!

kasbah:
Great, though my example name might not be the best. These SWR hook names seem to easily become pretty horrible. Should probably just be `useAllRepos`, `useUser`, `useUserRepos` etc. leaving out `Fetch`.

AbdulrhmnGhanem:
Ignore the use of SWR in the `uploadContext` anyway its structure will change as I change the logic for the uploading process(will work on this tomorrow as it's more relevant to a different issue #56 ).

frontend/src/contexts/UploadContext.js
```diff
@@ -105,9 +98,28 @@ export default function UploadContextProvider(props) {
...
   )
 }
+
+/**
+ * Fetch the files in the gitea repo associated with project being updated
+ * @param repo{string}
+ * @param shouldFetch{boolean}
+ * @returns {{isLoading: boolean, isError: boolean, files: any | Array | null}}
+ */
+const useFetchRemoteRepoFiles = (repo, shouldFetch) => {
```
kasbah, line 115: This name is still too long.

kasbah:
If there are any renames from `repo` to `project` left, let's get rid of them.

frontend/src/contexts/UploadContext.js
AbdulrhmnGhanem: > Ignore the use of SWR in the uploadContext anyway its structure will change as I change the logic for the uploading process(will work on this tomorrow as it's more relevant to a different pr #62).

This file was deleted as part of #62, see https://github.com/kitspace/kitspace-v2/pull/62/commits/2cd8561d5113cc73ae528f2c4dd11dfd716e01d8.

PR #58, opened 2021-01-02
Use polling when syncing repos

### LINKED ISSUE #56: Nginx timeout when syncing big repos

AbdulrhmnGhanem:
closes #56


nginx/kitspace.template
```diff
@@ -45,6 +45,33 @@ server {
...
+			add_header Content-Type text/plain;
+			add_header 'Access-Control-Allow-Origin' '@KITSPACE_URL';
+			add_header 'Access-Control-Allow-Methods' 'POST';
+			add_header 'Access-Control-Allow-Headers' 'cache-control, pragma, user-agent, content-type';
+			add_header 'Access-Control-Allow-Credentials' 'true' always;
+			return 200;
+		}
+		proxy_pass http://gitea:3000;
+	}
+
```
kasbah, line 74: Why not just increas the timeout for all gitea requests?

AbdulrhmnGhanem: If any other request took that long to resolve this means there's something wrong. The official docs advise not to use more than 75s, the default is 60s. Though for `/api/v2/repos/migrate` we know it's normal to take more than that.

AbdulrhmnGhanem:
Replaced by #62

AbdulrhmnGhanem:
#60 should get done first, the compilation takes too long using the original dockerfile.

PR #62, opened 2021-01-10
Redesign uploading process

AbdulrhmnGhanem:
- Instead of getting the content of the files from the `SessionStorage` it has to be accessed directly to be able to upload on a drop.
- Use the new utility to read the file content in `uploadFileToGiteaServer`.
- Replace the logic for loading files to `SessionStorage` and uploading the files using `commitFiles` Directly.
-  Replace the logic for loading files to `SessionStorage` and uploading the files by using `commitFiles` Directly.
- The`PreviewFiles` component now accepts the file as a prop.
- - Make `projects/update/[user]/[projectName]` responsible for fetching the remote file of the project being updated.
- The UploadContext.js is no longer needed.
- -For the update route, make the commit a two-step process:
   i.  upload on a drop,
   ii. commit on submit.
- Add validation to the project name, make sure there's no project with the same name that already exists.
- Add `MAX_FILE_SIZE` environment varible
   - It is used to configure both the frontend(dropzone will reject files larger than  `MAX_FILE_SIZE`) and backend(any request larger than  `MAX_FILE_SIZE` will return `413`)
 - Display failure notification uploading large files.
   - Make dropzone reject files larger than `MAX_FILE_SIZE` from the first place, though the backend will reject it anyway it's also good to prevent it from happening on the client-side.
   - Display failure notification uploading large files, this done by notifications not messages because displaying messages will require passing the props up to the parent components(context). It's easier to display a notification, the bundle will only be imported in case of failure.
 - Handle updating project name and uploading files at the same time.

The first commit related to this pr exclusively is 76d8bff. After merging the other branches into master, all the commits predating this one won't change.


kasbah:
Resolve conflicts please.

PR #66, opened 2021-02-06
Upload with tree structure

AbdulrhmnGhanem:
* Commit files with a tree structure.
   - Previously, committing files would commit all files to the root of the directory, ignoring the actual tree path.
   - Currently, uploading a folder either via Drag'nDrop or selector will commit files with the actual tree structure of the folder.
* Reject empty files.
   - Sometimes error message would appear, turned out trying to upload empty files breaks the encoding function.
* Return 404 page on deleting update page for a project.
* Fix folder selector.
   - Selecting Folder bypassed the DropZone and triggered the `onDrop` logic directly. Now, selecting a folder gets through the DropZone and its validation logic.


AbdulrhmnGhanem:
I split the pr into two as you asked, see #71.

PR #71, opened 2021-03-18
Migrate from SSG to SSR

AbdulrhmnGhanem:
SSG causes some problems because of the way `_app.js` set up.

* Add `UserExists` util function.
* SSR `mine` page.
* Migrate all pages: SSG -> SSR.
   - The way authentication works breaks SSG in production. SSR is the next best thing!
   - It worked before because Next re-run the build on each request in development, see [docs](https://nextjs.org/docs/basic-features/data-fetching#runs-on-every-request-in-development)
* Configure frontend logger for production.
   - Previously there was no output from when running in production.
* Remove `ref` from the `useDefaultBranchFiles`.
  - Gitea returns the files in the default branch if there's no ref provided.

### DIFF
--- a/frontend/src/pages/projects/update/[user]/[projectName].js
+++ b/frontend/src/pages/projects/update/[user]/[projectName].js
@@ -54,12 +54,8 @@ export const getServerSideProps = async ({ params, query }) => {
   }
 }
 
-<<<<<<< HEAD
-  const { repo: project, isLoading } = useRepo(fullName)
-=======
 const UpdateProject = ({ repo, repoFiles, isSynced, user, projectName, isNew }) => {
   const fullName = `${user}/${projectName}`
->>>>>>> 73b4d1e (SSR project update page.)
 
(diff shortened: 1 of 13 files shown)


frontend/src/pages/projects/update/[user]/[projectName].js
```diff
@@ -27,18 +32,38 @@ import { AuthContext } from '@contexts/AuthContext'
...
+    return { notFound: true }
+  }
+}
 
+<<<<<<< HEAD
   const { repo: project, isLoading } = useRepo(fullName)
+=======
+const UpdateProject = ({ repo, repoFiles, isSynced, user, projectName, isNew }) => {
+  const fullName = `${user}/${projectName}`
+>>>>>>> 73b4d1e (SSR project update page.)
```
Comment (kasbah) [inline frontend/src/pages/projects/update/[user]/[projectName].js:62]: Uh oh!

Comment (AbdulrhmnGhanem) [reply frontend/src/pages/projects/update/[user]/[projectName].js:62]: I will rename them

PR #73, opened 2021-03-18
Make default docker compose useful for frontend dev again

kasbah:
Not sure why this was running in production mode in for the development docker-compose.yml or why this wasn't caught on review. Also not sure how you have been running a frontend dev server @AbdulrhmnGhanem because the docker-compose.yml as it was wasn't very useful for that.


AbdulrhmnGhanem:
I changed it while testing SSR because SSG worked in development but didn't work in production. It got reset to the normal configuration in #71, which is the base for other PRs.

PR #74, opened 2021-03-19
Enable selecting key files(board, readme, BOM) from uploaded files

### LINKED ISSUE #70: Enable selecting key files(board, readme, BOM) from uploaded files
Users should be able to select/upload key files which was the role of the `kitspace.yaml` file.

image
image

AbdulrhmnGhanem:
- [x]  Port kitspace-v1 functionality
    - [x]  Info bar
    - [x]  Board showcase
    - [x]  Extra menu
        - [x]  Tracespace
        - [x]  IBoM
    - [x]  Order PCB
    - [x]  Buy parts
    - [x]  Readme
    - [x] IBOM page
- [x]  Integrate tree file preview
- [x]  Persist selected files on gitea

## Needed fixes
- [ ] Global refacotring.
- ~~[ ] Fix `1-click-bom` extension integration.~~.  #125.
- [x] Revalidate the content of tree preview after submission.
- [x] Reintegrate the drop zone on the project page.
- [x]  Allow only expected asset type to be selected(folder for gerbers, file for readme, etc).
- [x] "Update BOM file" button is in the wrong place (can we call these buttons "Edit" and make it obvious what they will edit?)
- [x] "Buy Parts" buttons are wrong colors
- [x] Drop zone improvements: 
     - [x] Drop zone and file tree are too close together
    -  [x] Remove bold style from "Drop", 
    -  ~~[ ] Drop zone should get rid of buttons, just use `<a>`~~,
       -  [x] use minimal sytle for dropzone buttons 
    -  [x] Stop mentioned "BOM", "README" etc. just `"Drop files here, <a>browse files</a> or <a>select a folder</a>"`
    -  [x] Need to differentiate between when we want one file (README, BOM) and when we want multiple files (PCB files). We don't need to be able to upload a folder when we are looking for one file. 
- [x] File tree 
    - [x] needs some text to say what it does "Select from previously uploaded files:"
    - [x] cursor should be "pointer" style, not text select, 
    - [x] needs a hover style too
    - [x] when you click on a file, the modal should close. for the gerbers one, it should select one folder. 
- [x] IBOM title bar is too tall, takes up too much space.
- [x] Tracespace icon color isn't correct.
- [x] Maximum width seems to be more limited than on kitspace.org.


AbdulrhmnGhanem:
Should this button be in its current position or below as the arrow indicates? I think moving makes sense because these two buttons' functionality depends on PCB files. 

image

Should I make any button that will modify the project visible only to the project owner?
We can change this later after implementing pull requests.

AbdulrhmnGhanem:
`info.has_interactive_bom` and `info.id` which are used in [v1](https://github.com/kitspace/kitspace/blob/adfda0f6ef9dbfebdcff455c0b199ca4354a5f74/src/page/page.jsx#L65) aren't in `info.json` file, is that right?

kasbah:
Yes makes sense to have it below and yes only show it for users with edit privileges. 

The interactive BOM setting isn't in info.json / kispace.yaml yet but we should put it there.

kasbah:
Probably would make most sense like this.  

image

kasbah:
On Second thought, it should go on the top. On the README you don't want to scroll to the bottom to find the button. The rest should go on the top for consistency. I have updated the Balsamiq sketch with this, and some other tweaks: https://balsamiq.cloud/sdxjss6/pbqwocj

AbdulrhmnGhanem:
The `static/images` isn't reachable in production, it's probably related to either Nginx or production vloumes.

kasbah:
Not too familiar with it myself but looks like it's generated from [this task](https://github.com/kitspace/kitspace/blob/2f2cd4599196a3ceb8e3f5ff62cc1c3a81811ada/tasks/makeIBOMJsx.js).

kasbah:
May be better to use [`dangerouslySetInnerHTML`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) than the conversion we are doing over there.

AbdulrhmnGhanem:
There's a `modules` field in `interactive_bom.json`, it's missing. I couldn't find how to generate it.

AbdulrhmnGhanem:
Do you know the source of this error
```sh
TypeError: Cannot read property 'trace' of null
    at interpolate (/app/node_modules/gerber-plotter/lib/_operate.js:385:12)
    at operate (/app/node_modules/gerber-plotter/lib/_operate.js:436:13)
    at Plotter._transform (/app/node_modules/gerber-plotter/lib/plotter.js:180:18)
    at Plotter.Transform._read (/app/node_modules/readable-stream/lib/_stream_transform.js:177:10)
    at Plotter.Transform._write (/app/node_modules/readable-stream/lib/_stream_transform.js:164:83)
    at doWrite (/app/node_modules/readable-stream/lib/_stream_writable.js:409:139)
    at writeOrBuffer (/app/node_modules/readable-stream/lib/_stream_writable.js:398:5)
    at Plotter.Writable.write (/app/node_modules/readable-stream/lib/_stream_writable.js:307:11)
    at Parser.ondata (/app/node_modules/readable-stream/lib/_stream_readable.js:681:20)
    at Parser.emit (events.js:315:20)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

It happens only on production in the processor.

kasbah:
Hm, looks like a problem with the tracespace lib. We'll need to upgrade to latest and see if it still happens.

AbdulrhmnGhanem:
> Hm, looks like a problem with the tracespace lib. We'll need to upgrade to latest and see if it still happens.

But the same version is used locally.
I pruned the staging server, the processor is back to normal. Though I don't know why this solved the issue.

kasbah:
May only be revealed by a particular board?

AbdulrhmnGhanem:
Gitea had only ulx3s which is the main testing board, and it never caused a problem.

AbdulrhmnGhanem:
Have you tested the new method for installing gitea, as part of the e2e? it's [failing](https://github.com/kitspace/kitspace-v2/runs/2410059728?check_suite_focus=true#step:7:220).

kasbah:
It worked on master: https://github.com/kitspace/kitspace-v2/runs/2346022529#step:7:1

kasbah:
I did docker stop, docker prune etc on https://master.staging.kitspace.dev and re-ran the job, deployment and e2e is working on master (ignoring flaky tests) so it must be about something you are changing on this branch.

docker-compose.yml
```diff
@@ -31,6 +36,7 @@ services:
       - /etc/localtime:/etc/localtime:ro
     depends_on:
       - postgres
+    command: /bin/bash --verbose -c "make build && cp gitea /app/gitea/ && /usr/bin/entrypoint"
```
kasbah, line 39: That could be a problem, that command should be in .override.yml

nginx/kitspace.template
```diff
@@ -104,10 +100,21 @@ server {
...
+		proxy_set_header    X-Forwarded-For               $proxy_add_x_forwarded_for;
+		proxy_set_header    X-Forwarded-Proto             $scheme;
+		add_header          'Access-Control-Allow-Origin' '*' always;
+		add_header          'Access-Control-Allow-Methods' 'GET, OPTIONS';
+
+		if ($request_method = OPTIONS) {
+			add_header Content-Length 0;
+			add_header Content-Type text/plain;
+			add_header 'Access-Control-Allow-Origin' '*';
+			add_header 'Access-Control-Allow-Methods' 'GET, OPTIONS';
```
kasbah, line 114: Something I was wondering for a while, why are we adding these headers twice?

AbdulrhmnGhanem:
Should we support `rst` rendering for the readme? If yes I will implement it as part of gitea. I will see if the output of custom renderers can be exposed as endpoints.

kasbah:
We will later but let's leave it for now.

frontend/Dockerfile.prod
```diff
@@ -1,4 +1,4 @@
-FROM node:10.16.0-alpine
+FROM node:14.16.1-alpine3.13
```
kasbah, line 1: Something I realized recently: we need to use less specific version (e.g. `node:14`) or we will never pull in security updates.

AbdulrhmnGhanem: Resolved here 1d6a27b.

AbdulrhmnGhanem:
Does the processor create a new file from `kitspace.yaml`? If no, I will make it do so. I think it's better to have this data as JSON for this scenario.
- After selecting the files, a new file `kitspace.yaml` will get committed to the repo.
- If the user tries to change the selection(e.g., changed the readme path), before committing the new selection, the file in the gitea repo has to be read to get other data fields(gerbers, 1-click-bom, etc...).
If the data is provided as json there will be no need to interact directly with the file in the gitea repo. This means no need to have something to deal with YAML in the frontend.

AbdulrhmnGhanem:
@kasbah, would you work on this?
> There's a modules field in interactive_bom.json, it's missing. I couldn't find how to generate it.

AbdulrhmnGhanem:
Please, take a look at staging, in case if you have comments on the file tree.

kasbah:
> There's a modules field in interactive_bom.json, it's missing. I couldn't find how to generate it.
I know as much as you about that one. Please try and figure it out.

kasbah:
File tree looks good, some things I noticed:

- [ ] "Update BOM file" button is in the wrong place (can we call these buttons "Edit" and make it obvious what they will edit?)
- [ ] "Buy Parts" buttons are wrong colors
- [ ] Drop zone and file tree are too close together
- [ ] Drop zone improvements: 
    -  [ ] Remove bold style from "Drop", 
    -  [ ] Drop zone should get rid of buttons, just use `<a>`, 
    -  [ ] Stop mentioned "BOM", "README" etc. just `"Drop files here, <a>browse files</a> or <a>select a folder</a>"`
    -  [ ] Need to differentiate between when we want one file (README, BOM) and when we want multiple "PCB files"
- [ ] File tree 
    - needs some text to say what it does "Select from previously uploaded files:"
    - cursor should be "normal" style not text select
    - needs a hover style too
    - when you click on a file, the modal should close

kasbah:
>>  There's a modules field in interactive_bom.json, it's missing. I couldn't find how to generate it.
>
> I know as much as you about that one. Please try and figure it out.

IBOM seems to be working without it perfectly though, no?

AbdulrhmnGhanem:
> IBOM seems to be working without it perfectly though, no?

Yes. Though if a board isn't supported by the IBOM, the IBOM icon will get displayed. Currently, it will return 404, is that ok?

kasbah:
No, we need to figure out how to remove it in those cases. Surely that's not related to `modules` in the JSON though?

AbdulrhmnGhanem:
> Surely that's not related to `modules` in the JSON though?

Fixed it in https://github.com/kitspace/kitspace-v2/pull/74/commits/e83f4f30556ca22b9221c52fd259db692708f6c2.

kasbah:
Oh, you solved it 10 days ago? Why were you asking me to take a look at it?

AbdulrhmnGhanem:
Uh, sorry! Wrong quote. I meant the `info.has_ibom`

> `info.has_interactive_bom` and `info.id` which are used in [v1](https://github.com/kitspace/kitspace/blob/adfda0f6ef9dbfebdcff455c0b199ca4354a5f74/src/page/page.jsx#L65) aren't in `info.json` file, is that right?

The `info.id` comes from the gitea API now. The problem is `has_ibom`.

kasbah:
Ah, I see. Can you use the `/status/` endpoint on the processor? I believe it will return `424` when it has failed, which, if I remember it correctly, currently includes not finding the files to process. 

https://github.com/kitspace/kitspace-v2/blob/1eda6b812555a216488f059e3d26ee73d9570e9c/processor/src/app.js#L84-L87

AbdulrhmnGhanem:
There's another problem with files in [omit-ibom-boards](https://github.com/kitspace/kitspace/blob/master/omit-ibom-boards.txt). I think this option should be added to the `kitspace.yaml`

frontend/src/utils/projectPage.js
```diff
@@ -0,0 +1,34 @@
...
+/**
+ *  
+ * @param {string} repoFullname 
+ * @returns{Promise<boolean>}
+ */
+export const hasInteractiveBom = async repoFullname => {
+  const res = await fetch(`${processorUrl}/files/${repoFullname}/HEAD/interactive_bom.json`)
+
+  return res.status !== 424
+}
```
kasbah, line 34: I think this should be:
```js
const res = await fetch(`${processorUrl}/status/${repoFullname}/HEAD/interactive_bom.json`).then(r => r.json())

return res.status === 'done'
```

Keeping in mind that it can also be `'in_progress'` or `'failed'`.

kasbah:
I've just documented the protocol [here](https://github.com/kitspace/kitspace-v2/blob/61fc219afe265635a0dc08da50cbeb0eb2393f7a/processor/README.md#protocol).

AbdulrhmnGhanem:
I am not sure I understand this, and why?

> Drop zone should get rid of buttons, just use <a>,

For now, I added `as='a'` which will render the button as `a`.

> Need to differentiate between when we want one file (README, BOM) and when we want multiple files (PCB files). We don't need to be able to upload a folder when we are looking for one file.

My original plan was to accept anything in the dropzone and make the actual selection by checking the file, and the submit button. I think your idea is better.

nginx/kitspace.template
AbdulrhmnGhanem: Yes, the preflight methods are handled separately from other methods.

kasbah:
Rendering them as links as buttons is about how they look, not the underlying element. The problem is they are grabbing too much attention, they should just look like links.

nginx/kitspace.template
kasbah: Not really though. I think these lines are redundant. We are already adding the headers above.

AbdulrhmnGhanem:
>  needs a hover style too

This gives a false sense that clicking on the file name will do something, should it open the file?
image

kasbah:
I think when you click on it it should just select that file and "submit" automatically.

nginx/kitspace.template
kasbah: Have you tried just adding these two plus `return 200;` ? The rest of the seem to be repeating what's above. 

https://github.com/kitspace/kitspace-v2/blob/25cd4f45ee59183bf5f3e41e09b0bc73781bc119/nginx/kitspace.template#L82-L83

AbdulrhmnGhanem:
The file update endpoint is broken even on the official testing instance https://try.gitea.io/api/swagger#/repository/repoUpdateFile

kasbah:
You know where to report that ;)

nginx/kitspace.template
AbdulrhmnGhanem: But this will overwrite the response for non-options requests.
Aside, this `200` should be `201`.

AbdulrhmnGhanem:
> The file update endpoint is broken even on the official testing instance https://try.gitea.io/api/swagger#/repository/repoUpdateFile

> You know where to report that ;)

My bad, I missed a parameter from the documentation, the file `sha` should be sent in the request body.

kasbah:
Noticed some more things:
- Tracespace icon color isn't correct
- Maximum width seems to be more limited than on kitspace.org, compare:


image

kitspace.org:
image

AbdulrhmnGhanem:
How does the processor detect the gerbers in [this](https://github.com/Jan--Henrik/Otter-Iron/blob/master/kitspace.yaml) project? There's a typo instead of `gerbers` it's `gebers`.

kasbah:
:sweat_smile:  
 
The way it works is it runs [whats-that-gerber](https://github.com/tracespace/tracespace/tree/master/packages/whats-that-gerber) on _all_ repo files when there is no `gerbers:` field. It then checks for duplicates and picks a single folder if there are duplicates.
 
 https://github.com/kitspace/kitspace-v2/blob/cdef55caf6537ba76d4ead61046fc7ab7bf65e31/processor/src/tasks/processGerbers/gerber_files.js#L4-L41
 
 Since we need this logic on both frontend and backend it might be time to make some re-usable modules. I would suggest to use [yarn workspaces](https://classic.yarnpkg.com/en/docs/workspaces/) for this, but the way we can make this work with Docker is a bit unclear to me.

AbdulrhmnGhanem:
I prefer keeping this in the processor; we can patch `kitspace-yaml.json` with the used assets, probably rename `kitspace-yaml.json` to something different not to confuse it with what we have now.

kasbah:
So we could do:

GET `processor.kitspace.test/[user]/[project]/HEAD/file-map.json`

Which could be e.g. 

```json
{
  "manufacturing": {
    "gerbers/my-board-F_Cu.gbr": {"type": "copper", "side": "top"},
    "gerbers/my-board-B_Cu.gbr": {"type": "copper", "side": "bottom"},
    "gerbers/my-board-notes.gbr": {"type": "drawing", "side": null}
  },
 "bom": "1-click-bom.csv",
 "readme": "README.md"
}
```
It could also be the KiCad (or Eagle file at some point):

```json
{
  "manufacturing": {
    "my-board.kicad_pcb":  {"type": "kicad", "side": null}
  },
 "bom": "bom.xlsx",
 "readme": "README.md"
}
```

At some point we'll probably want to include  an EDA field as well, since it is treated differently and used for different things. 

```json
{
  "manufacturing": {
    "my-board.kicad_pcb":  {"type": "kicad", "side": null}
  },
 "bom": "bom.xlsx",
 "readme": "README.md",
 "eda": {
     "type": "kicad",
     "pcb": "my-board.kicad_pcb",
     "sch": "my-board.sch"
  } 
}

**Comment by AbdulrhmnGhanem** [comment]
Great, I like this!

**Comment by AbdulrhmnGhanem** [comment]
I faced the issue with `file.path` in electron now in the CI. Oddly I am using Electron locally, and it works fine!

File: frontend/Dockerfile.dev [hunk 5]
```diff
@@ -2,6 +2,8 @@ FROM node:14-alpine
 
 ENV NODE_ENV=development
 
+RUN apk --no-cache add git
+
```
kasbah, line 6: What's that for?

frontend/src/utils/index.js
```diff
@@ -22,6 +28,65 @@ export const slugifiedNameFromFiles = files => {
...
+    const href = a.attr('href')
+    if (href.startsWith('/')) {
+      const rawUrl = `${giteaUrl}/${repo}/src${href}`
+      console.log(rawUrl)
+      a.attr('href', rawUrl)
+    }
+  })
+
+  return $.html()
+}
```
kasbah, line 88: I thought we would use the Gitea readme renderer.

nginx/kitspace.template
kasbah: You are still adding headers twice for `OPTIONS`, once on line 107-108 and once on lines 113-114.

processor/src/app.js
```diff
@@ -37,21 +37,6 @@ function createApp(repoDir = '/gitea-data/git/repositories') {
...
-    if (ALLOWED_CORS_DOMAINS.includes(origin)) {
-      res.header('Access-Control-Allow-Origin', origin)
-      res.header('Access-Control-Allow-Methods', 'GET,OPTIONS')
-      res.header('Access-Control-Allow-Headers', 'Content-Type')
-      res.header('Access-Control-Allow-Credentials', 'true')
-      return next()
-    }
-    return res.sendStatus(403)
-  })
-
```
kasbah, line 54: The env variable needs to be removed properly everywhere.

kasbah (changes requested):
I don't think this fully implements #70 yet, so it should be kept open. But we should stop having really long running branches so should try and merge.

For the future, I have been thinking about this: please try and find a cut-off much earlier and submit much smaller pull-requests. It will make review and merging much easier. Try and break the tasks into much smaller sub-tasks and submit pull requests for each of these. You can base subsequent pull-requests on previous branches as you continue work, then rebase them when the previous branch is merged in. 

E.g. for this one it could have been a sub-task to just get the board displaying on the project page.

frontend/Dockerfile.dev
AbdulrhmnGhanem: Installing `1-click-bom` https://github.com/kitspace/kitspace-v2/blob/e4e9ab98ed31d8bc61246ce5b69570dec681323c/frontend/package.json#L13

AbdulrhmnGhanem: Installing `1-click-bom` https://github.com/kitspace/kitspace-v2/blob/e4e9ab98ed31d8bc61246ce5b69570dec681323c/frontend/package.json#L13

kasbah: Can you add some comments please, so know we can remove it when we no longer have git dependencies.

frontend/src/utils/index.js
AbdulrhmnGhanem: Yes, I did so. This part only fixes the URLs.

processor/src/app.js
AbdulrhmnGhanem: Which env variable?

kasbah: ALLOWED_CORS_DOMAINS for the processor

frontend/src/components/Board/BoardShowcase.module.scss
```diff
@@ -0,0 +1,209 @@
...
+    @media (max-width: 499px) {
+      @content;
+    }
+  } @else if $point == mobile_menu {
+    @media (max-width: 590px) {
+      @content;
+    }
+  }
+}
+
```
kasbah, line 46: Why are these repeated everywhere? Can we not `@import` them?

frontend/Dockerfile.dev
AbdulrhmnGhanem: Whenever I install a new dependency I add the why to the commit, see https://github.com/kitspace/kitspace-v2/commit/bc5474254f25cf772d31f1e7a9ffa4939c298b7d

frontend/src/components/Board/BoardShowcase.module.scss
AbdulrhmnGhanem: Yes, all the styles need proper reorganization.

frontend/Dockerfile.dev
kasbah: Can you add comment to the Dockerfile please. We may move/rename the Dockerfile and the git comment will be hard to track down. We need a little reminder to remove the git install from the Dockerfile if we don't have git dependencies.

frontend/src/utils/index.js
kasbah: We should make an issue to do this in Gitea itself. Also, there's a `console.log` in there.

frontend/src/components/Board/BoardShowcase.module.scss
kasbah: Can you do that on this PR please.

frontend/Dockerfile.dev
AbdulrhmnGhanem: Resolved via d168558.

processor/src/app.js
AbdulrhmnGhanem: Resolved via 09f3f2b.

nginx/kitspace.template
AbdulrhmnGhanem: Resolved via 081a257.

nginx/kitspace.template
```diff
@@ -68,20 +68,14 @@ server {
...
-		add_header                 'Access-Control-Allow-Headers'     'cache-control, pragma, user-agent';
+		add_header                 'Access-Control-Allow-Headers'     'cache-control, pragma, user-agent, content-type';
 		proxy_cookie_path          ~^(.+)$                            '$1; domain=.@KITSPACE_DOMAIN';
 		proxy_set_header           Origin                             gitea.@KITSPACE_DOMAIN;
 		proxy_pass_request_headers on;
 		client_max_body_size       @MAX_FILE_SIZE;
 
 		if ($request_method = OPTIONS) {
-			add_header Content-Length 0;
-			add_header Content-Type text/plain;
```
kasbah, line 83: We don't need these two? Content-Lenght 0 and Content-Type text/plain?

AbdulrhmnGhanem: No, `204` do the same job.

frontend/src/components/Board/BoardShowcase.module.scss
AbdulrhmnGhanem: Resolved via e2043bc.

frontend/src/components/Board/BuyParts/MpnPopup.module.scss
```diff
@@ -0,0 +1,28 @@
...
+    padding-right: 10px;
+    align-items: flex-end;
+    flex-direction: column;
+  }
+  .octopartLinkContainer {
+    font-size: 10px;
+    justify-content: flex-start;
+    align-items: flex-start;
+  }
+}
```
kasbah, line 28: Now that these are .module.scss I think we can get rid of the outer classes, i.e. `.MpnPopup` in this case.

kasbah (changes requested):

frontend/src/utils/index.js
kasbah: Made an issue: https://github.com/kitspace/kitspace-v2/issues/128

frontend/src/pages/[username]/[projectName]/index.js
```diff
@@ -0,0 +1,413 @@
...
+import Readme from '@components/Board/Readme'
+import UploadModal from '@components/UploadModal'
+
+export const getServerSideProps = async ({ params, query, req }) => {
+  const processorUrl = process.env.KITSPACE_PROCESSOR_URL
+  const repoFullname = `${params.username}/${params.projectName}`
+  const assetsPath = `${processorUrl}/files/${repoFullname}/HEAD`
+
+  // Only the repo owner can upload files.
+  const hasUploadPermission = params.username === req?.session?.user?.username
```
Comment (kasbah) [inline frontend/src/pages/[username]/[projectName]/index.js:49]: I think we should be checking the Gitea API for if the user has write permission to this project.

frontend/src/pages/[username]/[projectName]/index.js
```diff
@@ -0,0 +1,413 @@
...
+import BoardExtraMenus from '@components/Board/BoardExtrasMenu'
+import OrderPCBs from '@components/Board/OrderPCBs'
+import BuyParts from '@components/Board/BuyParts/index'
+import InfoBar from '@components/Board/InfoBar'
+import Readme from '@components/Board/Readme'
+import UploadModal from '@components/UploadModal'
+
+export const getServerSideProps = async ({ params, query, req }) => {
+  const processorUrl = process.env.KITSPACE_PROCESSOR_URL
+  const repoFullname = `${params.username}/${params.projectName}`
```
Comment (kasbah) [inline frontend/src/pages/[username]/[projectName]/index.js:45]: This should also come from Gitea API, `project.full_name` I believe.

frontend/src/pages/[username]/[projectName]/index.js
```diff
@@ -0,0 +1,413 @@
...
+  // Set values of the form as the values of the project stored in the Gitea repo
+  useEffect(() => {
+    populate({ name, description }, true)
+  }, [])
+
+  useEffect(() => {
+    // Handle client side rendering for uploading permissions,
+    // `canUpload` previously relied on `hasUploadPermission` which is only provided in SSR mode.
+    if (!hasUploadPermission) {
+      setCanUpload(user?.username === owner && !previewOnly)
```
Comment (kasbah) [inline frontend/src/pages/[username]/[projectName]/index.js:212]: Here again, needs to check Gitea API if user has write permission.

frontend/src/components/Board/BuyParts/MpnPopup.module.scss
AbdulrhmnGhanem: Resolved via ade6a5d.

frontend/src/pages/[username]/[projectName]/index.js
Comment (AbdulrhmnGhanem) [reply frontend/src/pages/[username]/[projectName]/index.js]: Resolved via e998119.

frontend/src/pages/[username]/[projectName]/index.js
Comment (AbdulrhmnGhanem) [reply frontend/src/pages/[username]/[projectName]/index.js]: Resolved via e998119.

frontend/src/pages/[username]/[projectName]/index.js
Comment (AbdulrhmnGhanem) [reply frontend/src/pages/[username]/[projectName]/index.js:48]: I added a comment to clarify it and passed this to all the child components instead of constructing the name in each component.
https://github.com/kitspace/kitspace-v2/blob/e998119d55146baf0f7c7f65914cc4a259be48f4/frontend/src/pages/%5Busername%5D/%5BprojectName%5D/index.js#L46-L48

kasbah (approved):

Issue #38, opened 2020-12-22
Set up a Github action to run the cypress tests


kasbah:

AbdulrhmnGhanem:
This will be a little tricky because of the Gitea setup setp. Probably will need a puppeteer script to do this step.

kasbah:
Ah yeah, there's probably a way to get Gitea to finish the DB setup and create an admin user but I haven't figured it out yet.

kasbah:
```
gitea admin create-user --username user01 --password password --email user01@example.com --admin
```

AbdulrhmnGhanem:
So for the CI, my plan is:
- Build all the docker images and tag them with the commit sha that triggered the workflow,
- run the tests.
- if all the tests pass tag the images `latest` and upload them to `ghcr`

Are OK with that?

kasbah:
Sounds like a plan.

AbdulrhmnGhanem:
> ```
> gitea admin create-user --username user01 --password password --email user01@example.com --admin
> ```

This line only creates the admin, the setup process involves more steps, for example, the database must be ready before issuing this command. Couldn't find such a command in the gitea command line [page](https://docs.gitea.io/en-us/command-line/). I guess a puppeteer script would be the easiest way to do the headless setup.

kasbah:
I mean, the database should be ready. That confirmation step via the web interface always seemed superflous to me. 

The above command I noticed when trying out this algola demo: https://github.com/camandel/agola-demo. I don't think it has any database setup script. Not sure what's different. Maybe we need to switch to sqlite for these tests?

AbdulrhmnGhanem:
I added the command as the last `RUN` step in the Dockerfile, actually, this exact command is outdated and it should be `gitea admin user create --username username --password password --email example@example.com --admin` but it complained about the database.

kasbah:
Maybe Gitea actually has to be running? In the Vagrant script it waits for it: https://github.com/camandel/agola-demo/blob/6d6ffbf31fb126f758e4fb03d76fbdb384a20a2f/Vagrantfile#L55

kasbah:
Dockerfile seems like the wrong place anyway, we don't want to bake this into our images.

AbdulrhmnGhanem:
Initially thought of `entrypoint.sh` though this won't run the first time only.

I wrote a puppeteer script that configures gitea will make a PR now so you can take look.

AbdulrhmnGhanem:
[...]

PR #86, opened 2021-03-28
Simplify dev setup for gitea database

kasbah:
I just found out that the badly named `INSTALL_LOCK` env variable actually does what we wanted in #38. Additionally, after this I can run:

```
docker-compose run gitea \
gitea admin user create --username username --password password --email example@example.com --admin
```

So we can probably do something like that in the Github action too?


kasbah:
Even better, setting `SECRET_KEY` as we are supposed to anyway, achieves the same effect.

PR #89, opened 2021-03-31
Refactor GitHub actions

AbdulrhmnGhanem:
closes #83

PR #92, opened 2021-04-01
Refactor GitHub actions (2)

### LINKED ISSUE #83: Split github action `publish-docker-images` into build, test and publish phases
Split up this massive github action into seperate ones. Run build and test on all branches and PRs. Only publish master.

kasbah:
closes #83


kasbah:
@AbdulrhmnGhanem I think since we stopped running just on master/ghcr branches we are tagging every build from every branch with `latest` or am I missing something?

AbdulrhmnGhanem:
Every successful e2e, not build. We need to limit this to master

kasbah:
Any guess on these failures @AbdulrhmnGhanem ?

https://github.com/kitspace/kitspace-v2/runs/2247021262

Seems to be happening consistently on this branch now.

AbdulrhmnGhanem:
It times out. Previously the images were in the machine running the tests, but now it has to pull it, so it takes more than the specified timeout. Remove the timeout from this step and try again.

kasbah:
Which one is the timeout?

AbdulrhmnGhanem:
Install Gitea
https://github.com/kitspace/kitspace-v2/blob/1a7308d459ada199f65a226acd583cca5087f1ae/.github/workflows/publish-docker-images.yml#L151-L155

kasbah:
Ah, thanks. Shouldn't these containers be in the cache?

kasbah:
I think if that's the reason it would have failed on https://github.com/kitspace/kitspace-v2/pull/89 too, but it didn't. It must be something else I did on here.

AbdulrhmnGhanem:
Yes, it should usually it doesn't take more than 90s.

kasbah:
```
Head https://ghcr.io/v2/kitspace/processor/manifests/16529bc: unauthorized
```
Seems to be the problem. 

https://github.com/kitspace/kitspace-v2/pull/92/checks?check_run_id=2247367816#step:8:23

AbdulrhmnGhanem:
This failed too. Try pulling the images before running this step.

AbdulrhmnGhanem:
> Head https://ghcr.io/v2/kitspace/processor/manifests/16529bc: unauthorized

I have no idea what is this! These images are public I just pulled them locally without a problem.

kasbah:
I think I know, it's those `sed` calls. They are probably still looking for -v2 which I removed.

kasbah:
Or not https://github.com/kitspace/kitspace-v2/pull/92/checks?check_run_id=2247367816#step:5:24

kasbah:
Get the same thing locally. Processor isn't getting pushed to registry?

```
$ docker pull 'ghcr.io/kitspace/processor:16529bc'
Error response from daemon: Head "https://ghcr.io/v2/kitspace/processor/manifests/16529bc": unauthorized
```

AbdulrhmnGhanem:
Oh, you remove `v2`! All the new images are private now. I changed the visibility in the package setting for this repo. I believe this will fix it. If so, revert the timeout change.

AbdulrhmnGhanem:
I think removing the `yarn build` from the frontend makes loading the pages a bit slower. This might affect the tests, just a random guess.

Issue #77, opened 2021-03-26
Bit worried about the keys used for SWR


kasbah:
1. This one redundantly embeds the params into the endpoint and additionally into an object (where order and equality isn't guaranteed). 

https://github.com/kitspace/kitspace-v2/blob/2211fbf6908014f2209653ac085df005091f107f/frontend/src/hooks/Gitea.js#L86-L89

2. This one has a ternary, either using e.g.  `${giteaApiUrl}/users/null/repos` or most likely the bare `username`, as a key for the repos, which doesn't make a lot of sense to me.  

https://github.com/kitspace/kitspace-v2/blob/2211fbf6908014f2209653ac085df005091f107f/frontend/src/hooks/Gitea.js#L108-L111

kasbah:
I don't understand this one, even after looking at the docs: https://github.com/kitspace/kitspace-v2/blob/2211fbf6908014f2209653ac085df005091f107f/frontend/src/hooks/Gitea.js#L135

Why have a function that just returns the pre-computed `endpoint`?

kasbah:
Understanding 1. a bit better but I do think using an object there will cause trouble. Found this in the SWR docs:

> 
> ```js
> // Don’t do this! Deps will be changed on every render.
> useSWR(['/api/user', { id }], query)
> 
> // Instead, you should only pass “stable” values.
> useSWR(['/api/user', id], (url, id) => query(url, { id }))
> ```

https://swr.vercel.app/docs/arguments#passing-objects

AbdulrhmnGhanem:
> 1. This one redundantly embeds the params into the endpoint and additionally into an object (where order and equality aren't guaranteed).

Yes, after updating the endpoint I forgot to delete the params, this object is no longer needed.



> 2. This one has a ternary, either using e.g.  `${giteaApiUrl}/users/null/repos` or most likely the bare `username`, as a key for the repos, which doesn't make a lot of sense to me.

It will use this as a key `username ? endpoint : null`:
* if the `username` is `undefined` then it will evaluate to `null` which will prevent SWR from performing the request.
* Otherwise it will use the `endpoint` as a key, which contains the full repo name(username, and repo name)
Anyways I added this condition to get around a limitation that no longer exists; will remove it.

> I don't understand this one, even after looking at the docs:
> 
> https://github.com/kitspace/kitspace-v2/blob/2211fbf6908014f2209653ac085df005091f107f/frontend/src/hooks/Gitea.js#L131-L135
> 
[...]

PR #93, opened 2021-04-01
Clean gitea hooks

### LINKED ISSUE #85: `fetcher` can be improved
Including `initOps` confused me, the name is not clear ("initial" but then I would actually pass them to `useSWR(x, initOps)`?) Also OPS generally means "operations" and OPTS means "options". Also the documentation is wrong:

https://github.com/kitspace/kitspace-v2/blob/2211fbf6908014f2209653ac085df005091f107f/frontend/src/hooks/Gitea.js#L7

Anyway, that's a lot of words about one function argument. ~I think we should remove it.~

AbdulrhmnGhanem:
closes #85 #77

### DIFF
--- a/frontend/src/hooks/Gitea.js
+++ b/frontend/src/hooks/Gitea.js
@@ -65,6 +65,37 @@ export const useRepo = (fullname, swrOpts = {}) => {
 export const useAllRepos = (swrOpts = {}) => useSearchRepos(null, swrOpts)
 
 /**
+ * A hook to search all repos
+ * @param sort{string}
+ * @param order{string}
+ * @param swrOpts{Object}
+ * @param q{string=}: search query, leave undefined to return all repos
+ * @returns {{repos: [Object], IsLoading: boolean, IsError: boolean, mutate: function}}
+ */
+export const useSearchRepos = (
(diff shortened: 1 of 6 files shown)


frontend/src/hooks/Gitea.js
```diff
@@ -106,12 +72,7 @@ export const useSearchRepos = (
  */
 export const useUserRepos = (username, swrOpts = {}) => {
   const endpoint = `${giteaApiUrl}/users/${username}/repos`
-
-  const { data, error, mutate } = useSWR(
-    username ? endpoint : null,
-    fetcher,
-    swrOpts,
-  )
+  const { data, error, mutate } = useSWR(endpoint, fetcher, swrOpts)
```
kasbah, line 75: So this will end up fetching `/users/null/repos` when you use `useAllRepos` no?

AbdulrhmnGhanem: `useAllRepos` relies on `useSearch` not `useUserRepo`. Actually, `useAllRepos` isn't used anywhere

AbdulrhmnGhanem:
Squash  this

PR #98, opened 2021-04-05
.dev auto-deploys

kasbah:
I'm switching the auto-deploys domain to .dev because we could run into cookie issues when we finally deploy to the apex domain on .org. Additionally I've created two separate auto-deploys for our own branches. So the master branch deploy is now:

https://master.staging.kitspace.dev (hopefully we will see something here when this is merged) 

@AbdulrhmnGhanem you can push whatever you want to the `abdo-dev` branch and this should be deployed to:

https://abdo-dev.staging.kitspace.dev

I've added your public ssh key to authorized_keys so you should be able to login with `ubuntu@abdo-dev.staging.kitspace.dev`  when there are any issues (which there probably will be occasionally). 

My own branch is `kaspar-dev` which is deployed to: https://kaspar-dev.staging.kitspace.dev. We should make the e2e end tests run on an auto-deploy domain too.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -95,7 +95,7 @@ jobs:
           cache-to: type=local,dest=/tmp/.buildx-cache
 
   deploy:
-    if: ${{ github.ref == 'refs/heads/dev' || github.ref == 'refs/heads/master' }}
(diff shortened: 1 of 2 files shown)


AbdulrhmnGhanem:
Oh wow, great!

PR #104, opened 2021-04-26
Fix flaky tests

AbdulrhmnGhanem:


AbdulrhmnGhanem:
The last 6 runs got stable with only 2 flaky tests, so I focus on those.
image
Also, I think if we are going to consider flaky tests a failure case, then we should remove `retries`.

AbdulrhmnGhanem:
> Also, I think if we are going to consider flaky tests a failure case, then we should remove `retries`.

@kasbah, probably you haven't noticed this.

kasbah:
Remove `retries` from where? Not sure what you are talking about or what I would have been able to notice.

AbdulrhmnGhanem:
> Remove `retries` from where? Not sure what you are talking about or what I would have been able to notice.

Cypress retries, here:
https://github.com/kitspace/kitspace-v2/blob/7f8629e06ea05fb2865c1785a1dba8167ef17983/frontend/cypress.json#L10-L13

A test marked `flaky` by cypress when it runs it, it fails, it retries it again as many times as specified in `cypress.json` if all retries fail then this is marked as a failure not `flaky`.

kasbah:
Ok, seems like we should still have some retries though. Without retries, how would it identify a flaky test? Anyway, I'll let you handle it.

kasbah:
You think this is ready?

AbdulrhmnGhanem:
> You think this is ready?

Yes. Though I think we shouldn't consider flaky tests a failure in the future. This took too much time and effort yet providing little value.

frontend/cypress/integration/updateProject.spec.js
```diff
@@ -1,136 +1,5 @@
...
-    cy.get('[data-cy=update-form-submit]').click()
-
-    // Should redirect to the new update page
-    cy.url().should('contain', `/${username}/${newName}`)
-
-    // `example3.png` should be in the preview
-    cy.get('[data-cy=file-name]').contains('example3.png')
-  })
-})
-
```
kasbah, line 133: How come this whole test is being deleted?

kasbah:
> Yes. Though I think we shouldn't consider flaky tests a failure in the future. This took too much time and effort yet providing little value.

I disagree, we need to be able to trust the results from our test suite. Having a flaky test is worse than no test.

frontend/cypress/integration/updateProject.spec.js
AbdulrhmnGhanem: This is a mistake! I looked at the history for this file, I deleted it in an unrelated commit to this file, will roll that part back.

.github/workflows/docker.yml
```diff
@@ -171,11 +171,6 @@ jobs:
     runs-on: ubuntu-20.04
     needs:
       - build
-    strategy:
-      fail-fast: false
-      matrix:
-        # run copies of the current job in parallel
-        containers: [1, 2, 3, 4]
```
kasbah, line 178: Why are you getting rid of the parallel testing? Seems to speed things up quite a bit no?

AbdulrhmnGhanem: It seems it increases the flakiness of the test, probably due to the start-up time of the frontend. Even looking at the cypress dashboard didn't help, the recording starts at random steps. I leave this for a different pr.

PR #106, opened 2021-05-03
Add `multi` support to processor

kasbah:


AbdulrhmnGhanem:
So for the multi `[username]/[projectName]/[subProject].js` page is needed in the frontend?

kasbah:
Yes, (except for the `.js` bit). One thing we could do is provide `/files/[user]/[project]/HEAD/kitspace-yaml.json` which is just the parsed YAML in JSON format. That way you can list all the multi projects from the keys there, if that helps. I am open to suggestions on what you need for the frontend though.

AbdulrhmnGhanem:
> One thing we could do is provide `/files/[user]/[project]/HEAD/kitspace-yaml.json` which is just the parsed YAML in JSON format. T

Yes, https://github.com/kitspace/kitspace-v2/pull/74#issuecomment-830773939 :sweat_smile:

kasbah:
Ah, that's where I got the idea :D 

Any comments on the code itself? I'll merge if you don't see any issues.

AbdulrhmnGhanem:
> Any comments on the code itself? I'll merge if you don't see any issues.
No, looks good.

AbdulrhmnGhanem (approved):

PR #117, opened 2021-05-28
Use eslint

kasbah:


kasbah:
@AbdulrhmnGhanem do you know if we can get eslint to pay attention to the `noinspection` comments? Or the reverse, make your intelliJ respect `eslint-disable-next-line`.

AbdulrhmnGhanem:
I am no longer using IntelliJ, feel free to delete these lines and depend
on eslint alone.

frontend/src/components/Board/BuyParts/Bom.js
```diff
@@ -10,6 +11,7 @@ const Bom = ({ length, parts, tsv }) => {
   const [collapsed, setCollapsed] = useState(true)
 
   useEffect(() => {
+    // eslint-disable-next-line no-shadow
     const diff = length - 7
     setDiff(diff)
```
kasbah: What's being shadowed? I don't think we should make exceptions for this rule, we should either rename variables or decide we don't want this rule.

frontend/src/components/Board/BuyParts/DirectStores.js
```diff
@@ -0,0 +1,136 @@
...
+      if (code === 'GB') {
+        return 'UK'
+      }
+      if (usedCountryCodes.indexOf(code) < 0) {
+        return 'Other'
+      }
+      return code
+    } catch (err) {
+      // eslint-disable-next-line no-console
+      console.error(err)
```
kasbah, line 42: I think we should make global exceptions for `console.error` and `console.warn`.

frontend/src/components/Board/BuyParts/DirectStores.js
```diff
@@ -0,0 +1,136 @@
...
+    setDigikeyParts(getParts('Digikey'))
+    setFarnellParts(getParts('Farnell'))
+    setNewarkParts(getParts('Newark'))
+  }, [])
+
+  const tildeDelimiter = part => `${part.sku}~${part.quantity}`
+
+  const digikeyPartRenderer = (part, index) => {
+    // eslint-disable-next-line no-param-reassign
+    index += 1
```
kasbah, line 61: Do we want this rule? I always re-assign arguments.

kasbah (commented):
I realize I am reviewing prematurely. I think `eslint-disable` should really be a rare exception. Either disable the rule globally or fix the code. There can be some rare exceptions where the use of `eslint-disable` makes sense.

AbdulrhmnGhanem:
I am afraid to touch these two functions :laughing: 
https://github.com/kitspace/kitspace-v2/blob/2bea96d783a3308128f1de38c2719943c4614751/frontend/src/components/Board/BuyParts/index.js#L22-L33

https://github.com/kitspace/kitspace-v2/blob/2bea96d783a3308128f1de38c2719943c4614751/frontend/src/components/Board/BuyParts/TsvTable.js#L39-L48

frontend/src/components/Board/Readme.js
```diff
@@ -5,10 +5,7 @@ import _JSXStyle from 'styled-jsx/style'
 
 const Readme = ({ renderedReadme }) => (
   <div id="readme">
-    <div
-      // eslint-disable-next-line react/no-danger
-      dangerouslySetInnerHTML={{ __html: renderedReadme }}
-    />
+    <div dangerouslySetInnerHTML={{ __html: renderedReadme }} />
```
kasbah: I think this one is the one exception where I would suggest keeping the rule and using the `eslint-disable`, even though we use it in three places. We really need to know what we are doing with `dangerouslySetInnerHTML` and should avoid it if possible. We should have that extra check from eslint if someone tries to use it somewhere else.

frontend/src/components/Board/OrderPCBs.js
```diff
@@ -12,8 +12,7 @@ const OrderPCBs = ({ zipUrl, boardSpecs }) => {
   const jlcpcbUrl = `https://cart.jlcpcb.com/quote?fileurl=${zipUrl}&from=kitspace`
 
   // TODO
-  // eslint-disable-next-line no-unused-vars
-  const trackClick = vendor => e => {
+  const trackClick = () => () => {
     // window.plausible('Order PCBs', {
     //   props: { project: this.props.project, vendor },
     // })
```
kasbah: Would undo getting rid of `vendor` here too, made more sense before. By the way, I think we should change this to just check if `window.plausible` exists all the time, and no run it if it doesn't.

frontend/src/components/Board/Readme.js
AbdulrhmnGhanem: Resolved via 9b06911.

kasbah:
You force pushed over my commit: https://github.com/kitspace/kitspace-v2/commit/3d8ebf04b7e053d24a72d73ffb6aadce75e10745

AbdulrhmnGhanem:
> You force pushed over my commit: [3d8ebf0](https://github.com/kitspace/kitspace-v2/commit/3d8ebf04b7e053d24a72d73ffb6aadce75e10745). I think using `--force-with-lease` would have prevented that, please default to using `--force-with-lease`.

I have done so.

PR #120, opened 2021-05-28
Fix issues with file upload "path" property

kasbah:


AbdulrhmnGhanem:
Could you elaborate on  "issues"? I noticed some problems with dropping multiple files inside the project page, is it related to that?

frontend/src/utils/giteaInternalApi.js
```diff
@@ -90,10 +90,19 @@ export const commitInitialFiles = async ({
...
-  )
+  let filePaths = files.map(file => {
+    let filePath = file.path
+    if (!filePath) {
+      console.warn(
+        'File object in commitInitialFiles does not have a "path" property. Using "name" instead:',
+        file.name,
+      )
+      filePath = file.name
+    }
```
kasbah, line 101: This is the main change. We can't guarantee that `file.path` is there and not an empty string. Right now the only place where we saw `file.path === ""` was in Cypress on Electron (hence we saw the failures on CI) but we will need to do some more checks on different browsers. I checked unstable and beta chrome and it was fine, so I'm not sure what's happening with this API -- react-dropzone is also being very clever and filling it in when it can...

Anyway, I don't think it's related to any issue you saw with uploading multiple files.

frontend/cypress/support/index.js
```diff
@@ -1,3 +1,6 @@
 import 'cypress-fail-fast'
+import installLogsCollector from 'cypress-terminal-report/src/installLogsCollector.js'
+
+installLogsCollector()
 
 import './commands'
```
AbdulrhmnGhanem, line 6: I think eslint will complain about the import at line 6, keep the imports at top

AbdulrhmnGhanem (approved):

AbdulrhmnGhanem (approved):

PR #122, opened 2021-05-28
Add a user menu

kasbah:


AbdulrhmnGhanem:
## closed
image
# open
image
# small screen
image

AbdulrhmnGhanem:
I guess the implementation for the `/settings` page will be in a different PR, right?

.env.example
```diff
@@ -17,10 +17,6 @@ CYPRESS_GITEA_ADMIN_USERNAME=__admin
 CYPRESS_GITEA_ADMIN_PASSWORD=123456
 # Abort the rest of Cypress tests on first failure
 FAIL_FAST_ENABLED=true
-# Incremental static regeneration period(in seconds) for `/`
-INDEX_ISR_INTERVAL=3600
-# Incremental static regeneration period(in seconds) for `/projects/[user]`
-USER_ISR_INTERVAL=86400
```
kasbah, line 23: Why are these being removed? Are they not being used anymore?

AbdulrhmnGhanem:
> For that I think we need a clearer separation between the user menu and the rest of the options.

like this?
image

kasbah:
Yes, but that line is pretty ugly. I think there is a separator in Semantic UI that you can use. Also, I think the user stuff should be up top in the user menu.

frontend/src/components/NavBar/UserMenu.js
```diff
@@ -0,0 +1,77 @@
...
+import Link from 'next/link'
+import Image from 'next/image'
+import { useRouter } from 'next/router'
+
+import styles from './UserMenu.module.scss'
+import { AuthContext } from '@contexts/AuthContext'
+
+export const UserDropDownMenu = () => {
+  const { user } = useContext(AuthContext)
+  const [open, setOpen] = useState(false)
```
kasbah, line 12:
```suggestion
  const [isOpen, setOpen] = useState(false)
```

AbdulrhmnGhanem:
I increased the height. Though increasing the width to make as the pink rectangle will distort the image

image

PR #126, opened 2021-05-29
Add input file info to processor json endpoints

kasbah:
As discussed in https://github.com/kitspace/kitspace-v2/pull/74#issuecomment-850841145

### DIFF
--- a/frontend/src/components/Board/BuyParts/index.js
+++ b/frontend/src/components/Board/BuyParts/index.js
@@ -8,7 +8,7 @@ import InstallPrompt, { install1ClickBOM } from './InstallPrompt'
 import DirectStores from './DirectStores'
 import styles from './index.module.scss'
 
-const BuyParts = ({ project, lines, parts }) => {
+const BuyParts = ({ projectFullName, lines, parts }) => {
   const [extensionPresence, setExtensionPresence] = useState('unknown')
   // it's needed to fix the extension integration.
   // eslint-disable-next-line no-unused-vars
@@ -63,7 +63,7 @@ const BuyParts = ({ project, lines, parts }) => {
               setBuyParts(retailer => {
                 window.plausible('Buy Parts', {
                   props: {
-                    project,
+                    project: projectFullName,
                     vendor: retailer,
                     multiplier: mult,
                   },
@@ -91,7 +91,7 @@ const BuyParts = ({ project, lines, parts }) => {
       },
       false,
     )
-  }, [mult, project])
+  }, [mult, projectFullName])
 
   useEffect(() => {
     const multi = buyMultiplier
@@ -282,7 +282,7 @@ const StoreIcon = ({ retailer, disabled }) => {
 }
 
 BuyParts.propTypes = {
-  project: string.isRequired,
+  projectFullName: string.isRequired,
   lines: array.isRequired,
   parts: array.isRequired,
 }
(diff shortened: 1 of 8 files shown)


kasbah:
@AbdulrhmnGhanem I decided to change tthe plan a bit to fit with what the processor is already doing. This way I'm not adding new JSON files but just renaming some and adjusting the fields:

- What was `info.json` is now `bom-info.json`. It no longer has a `"site"` field (were you using this?) but still has the `"bom"` field. It now also has a `"inputFile"` field which is the path of the input BOM file (e.g. csv or xlsx etc.) 
- What was `zip-info.json` is now called `gerber-info.json` and has all the same fields but also an `"inputFiles"` field with the structure as we discussed previously

```json
{
  "inputFiles": {
    "gerbers/my-board-F_Cu.gbr": {"type": "copper", "side": "top"},
    "gerbers/my-board-B_Cu.gbr": {"type": "copper", "side": "bottom"},
    "gerbers/my-board-notes.gbr": {"type": "drawing", "side": null}
  }
}
```
or
```json
{
  "inputFiles": {
    "my-board.kicad_pcb":  {"type": "kicad", "side": null}
  }
}
```

The readme doesn't really fit well into the schema, the processor doesn't really care about the readme. I think we should come up with something else there.

AbdulrhmnGhanem:
Currently, `kitspace-yaml.json` is used for showing the selected assets in the file tree and updating `kitsapce. yaml` when selecting a different asset.

Why are the names of the files different than the fields in the `kitspace.yaml`? and splitting the `manufacturing` which was `gerbers(as single dir)` into multiple files?

kasbah:
The way the processor works makes it much harder to put everything into one JSON endpoint. We already have these endpoints so I put the info in there and renamed them. This is basically what I proposed on https://github.com/kitspace/kitspace-v2/pull/74#issuecomment-850841145 but the endpoints are separate. The information is the same. (Except for the readme, not sure what to do with that). 

The names are different to the kitspace.yaml because it's the result of a number of different factors. It's just telling you what files were actually read to get the output, hence `inputFiles`. Kitspace.yaml has some bearing on this, but both `gerbers` and `eda` fields will take an effect so it would be wrong to use those names.

I also don't think it's splitting "manufacturing" or "gerbers" into multiple files. It's now all in `gerber-info.json`.

AbdulrhmnGhanem:
>   * It no longer has a `"site"` field (were you using this?) but still has the `"bom"` field.

It's used here 
image

kasbah:
You can grab it from `kitspace-yaml.json` though. It just didn't make sense to put it in the `bom-info.json` any more.

AbdulrhmnGhanem:
These lines needs update

https://github.com/kitspace/kitspace-v2/blob/0958323bea9afa46313959cd63ccecd79e25ad4d/frontend/src/utils/projectPage.js#L14
https://github.com/kitspace/kitspace-v2/blob/0958323bea9afa46313959cd63ccecd79e25ad4d/frontend/src/utils/projectPage.js#L24

AbdulrhmnGhanem (approved):

kasbah:
I'll update the frontend on this branch as well. I'll wait for #74 to be merged in first. 

Not totally happy with this API and I can see how it's not ideal from the frontend perspective. Could you maybe sketch out what would be the best API you could imagine, from working on the frontend, forgetting everything we have already? We could think about wrapping everything we have in a single GraphQL endpoint for instance, to simplify things on the frontend side. What would that look like according to you @AbdulrhmnGhanem?

AbdulrhmnGhanem:
I never used GraphQL before. I will think of the design of the API.

Mainly, it should get all the data in `getServerSideProps` for the project page as a single endpoint.
https://github.com/kitspace/kitspace-v2/blob/0958323bea9afa46313959cd63ccecd79e25ad4d/frontend/src/pages/%5Busername%5D/%5BprojectName%5D/index.js#L43

PR #132, opened 2021-06-07
Browser matrix

AbdulrhmnGhanem:
Run integration tests on Chrome, Firefox, and Edge.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -183,7 +183,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        browsers: [chrome, firefox, edge]
+        browsers: [chrome]
 
     steps:
       - name: Check out the repo
@@ -242,7 +242,7 @@ jobs:
           wait-on-timeout: 120         
           working-directory: frontend
           record: true
-          command: yarn e2e --record --browser ${{ matrix.browser }}
+          browser: ${{ matrix.browsers }}
  
 
   publish:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -232,12 +232,14 @@ jobs:
           GITEA_DEPLOY_IMAGE_TAG: ':${{ steps.sha.outputs.gitea_sha_short }}'
         with:
           start: docker-compose -f ../docker-compose.yml -f ../docker-compose.deploy.yml up
-          command: yarn e2e --record --browser chrome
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yml
```diff
@@ -232,12 +247,14 @@ jobs:
...
-          command: yarn e2e --record --browser chrome
-          working-directory: frontend
           # wait for 2 minutes for the server to respond
           wait-on: 'http://kitspace.test:3000'
           # This value is in seconds
-          wait-on-timeout: 120
+          wait-on-timeout: 120         
+          working-directory: frontend
+          record: true
+          browser: ${{ matrix.browsers }}
```
kasbah, line 256: Why did you change this to use `command` before? I was non-command before then you changed it to be command and now back again. Just trying to understand...

AbdulrhmnGhanem: If the `command` key is present, the run options e.g., `record`, and `browser` are ignored.
In 6356f81088aad69abb363e3adf685d4f700fb1c4, I thought there was a problem with the run options, turned out I used `matrix.browser` instead of `matrix.browsers`.

AbdulrhmnGhanem:
You can squash this, just split it into multiple commits to see if any specific step will cause errors.

.github/workflows/docker.yml
```diff
@@ -180,6 +180,10 @@ jobs:
     runs-on: ubuntu-20.04
     needs:
       - build
+    strategy:
+      fail-fast: false
+      matrix:
+        browsers: [chrome, firefox, edge]
```
kasbah, line 186: I think we should run it in electron too. Since in practice that surfaced some potential bugs already.

kasbah:
While you are looking at tests. Can we reduce the flakiness test to a warning for now until we can completely fix the flakiness?

AbdulrhmnGhanem:
> While you are looking at tests. Can we reduce the flakiness test to a warning for now until we can completely fix the flakiness?

I don't have managing the dashboard, you can configure it as shown [here](https://docs.cypress.io/guides/dashboard/flaky-test-management#GitHub).

.github/workflows/docker.yml
AbdulrhmnGhanem: Resolved via 6c2ffac.

kasbah:
Thanks, I've disabled it. I disabled the pull-request comments as well.

PR #139, opened 2021-06-15
Upgrade to NextJS 11

kasbah:
https://nextjs.org/blog/next-11

Biggest change for us is built-in eslint config. Enabling it we fail some lint errors again, can you take this over @AbdulrhmnGhanem ?


kasbah:
Should take advanate for `next/image` as well: https://nextjs.org/blog/next-11#automatic-size-detection-local-images

AbdulrhmnGhanem:
There's a problem with the eslint for next and babel  https://github.com/vercel/next.js/issues/26127.

kasbah:
I think this is some upgrade issue with babel. I believe the work-around is to `rm -rf node_modules` and install again.

AbdulrhmnGhanem:
> `rm -rf node_modules` and install again.

It didn't fix it. Anyways this is a problem with eslint for the editor extension, works fine as expected from CLI.

frontend/src/components/Board/OrderPCBs.js
```diff
@@ -1,9 +1,12 @@
 import React from 'react'
+import Image from 'next/image'
 import { object, string } from 'prop-types'
 import { Icon } from 'semantic-ui-react'
 
 import styles from './OrderPCBs.module.scss'
 
+const retailerLogosDimensions = { height: 30, width: 100 }
+
```
kasbah, line 9: I believe you can get the dimensions automatically via `import`. 

https://nextjs.org/blog/next-11#automatic-size-detection-local-images

AbdulrhmnGhanem:
Do you see why the frontend can't start?

frontend/src/components/Board/OrderPCBs.js
AbdulrhmnGhanem: They don't have the same size, I guess the point of using `Image` over `img` is the optimization(resize).

frontend/src/pages/_app.js
```diff
@@ -1,3 +1,4 @@
+import '../wydr'
```
kasbah, line 1: The file should be renamed to `wdyr`

frontend/src/pages/search.js
```diff
@@ -40,9 +40,10 @@ const Search = ({ repos, q }) => {
 
   useEffect(() => {
     setQuery(isValid ? form.query : '')
-  }, [form])
+  }, [form, isValid])
 
   useEffect(() => {
+    /* eslint-disable react-hooks/exhaustive-deps */
```
kasbah, line 46: Can you explain in a comment why it doesn't apply here?

AbdulrhmnGhanem: Resolved via eaf3263.

frontend/src/pages/login.js
```diff
@@ -37,8 +37,18 @@ const Login = () => {
...
+                render: function SignUpTab() {
+                  return <SignUpForm />
+                },
+              },
+              {
+                menuItem: 'Log in',
+                render: function SignInTab() {
+                  return <SignInForm />
+                },
+              },
```
kasbah, line 51: Why is that being changed? Seems clearer and more compact before.

AbdulrhmnGhanem: There's a rule that all components should have a display name. This is the error when using anonymous functions
```
error  Component definition is missing display name  react/display-name
```

frontend/src/pages/search.js
kasbah: You replied to the wrong comment. I've marked the other one as resolved.

frontend/src/pages/_app.js
AbdulrhmnGhanem: Resolved via eaf3263.

frontend/src/pages/login.js
kasbah: Wait a minute, can't this just be: `render: SignUpform,` and `render: SignInForm,`?

frontend/src/pages/search.js
kasbah: Great, I think `eslint-disable` will disable it for the rest of the file though.

frontend/src/pages/login.js
AbdulrhmnGhanem: I tried this the first time working on this page and it doesn't work.
image
and the attributes of the `pane` prop
image

frontend/src/pages/search.js
AbdulrhmnGhanem: The  `/**/` works per block, removing a dependency from this hook
https://github.com/kitspace/kitspace-v2/blob/69e37ae343b45e852b5f9b9ac401938cc3655c15/frontend/src/pages/search.js#L41-L46
 which is above this block, result in this error

image

frontend/src/pages/login.js
kasbah: Those docs don't really explain why it wouldn't work. What actually happened when you tried it? It's not that important anyway, just leave it.

AbdulrhmnGhanem: > Those docs don't really explain why it wouldn't work

The `render` value should be a function that returns a component, not a component.

> What actually happened when you tried it?
Error with calling useEffect outside of React component.

AbdulrhmnGhanem:
I will kill the edge e2e, the download is taking more than 50 mins something is wrong with the provider.

frontend/package.json
```diff
@@ -30,41 +28,41 @@
...
     "eslint-config-prettier": "8.3.0",
     "eslint-import-resolver-alias": "^1.1.2",
     "eslint-plugin-import": "^2.23.4",
-    "eslint-plugin-jsx-a11y": "^6.4.1",
     "eslint-plugin-prettier": "3.4.0",
-    "eslint-plugin-react": "^7.21.5",
-    "eslint-plugin-react-hooks": "^1.7.0",
     "faker": "^5.1.0",
     "lint-staged": ">=10",
+    "mocha": "^3.5.0",
```
kasbah, line 65: Where is this coming from?

AbdulrhmnGhanem: cypress-terminal-report

kasbah: But nothing about that dependency changed and `mocha` is not a peer dependency of it. Why is it being added to our devDependencies?

AbdulrhmnGhanem: It's a peer dependency, you can remove it and the waring of missing peer dependency will appear when installing the packages.

kasbah: It doesn't seem to be a peer dependency in the most recent version at least. Even if we get a warning about peer dependencies it's not a good enough reason to add it to our devDependencies if we are not using mocha otherwise.

AbdulrhmnGhanem: Resolved via eadb08d.

frontend/src/components/Board/BuyParts/index.js
```diff
@@ -266,6 +266,9 @@ const RetailerButton = ({
 const StoreIcon = ({ retailer, disabled }) => {
   const imgHref = `/static/images/${retailer}${disabled ? '-grey' : ''}.ico`
   return (
+    // Styling this as a Next Image is unnecessarily complicated, also all next optimizations aren't that useful
```
kasbah, line 269: "all next optimizations..." isn't grammatically correct, I suggest "the next optimizations..."

kasbah: Comment is too long now :D

AbdulrhmnGhanem: Resolved via df88c9b.

frontend/src/components/NavBar.js
```diff
@@ -23,15 +25,30 @@ const NavBar = () => {
...
+        height={46}
+        objectFit="scale-down"
+        className={styles.logoImg}
+        src={logoSrc}
+        alt="logo"
+      />
+    </a>
+  )
+})
+
```
kasbah, line 42: Huh? forwardRef? What's going on here? Why not just use `Link` instead of `a`?

frontend/src/pages/[username]/[projectName]/IBOM.js
```diff
@@ -74,40 +63,59 @@ const IBOM = ({ html, pcbData }) => {
...
+          return
+        }}
+      />
+      <Script>{initScript}</Script>
+      <NextHead>
+        {/* Styles for IBOM targets uses non pure selectors, e.g., `:root`, importing its style as
+         * a  module doesn't work. The other option was to add the style to `_app.scss`
+         * which will load on visiting any page.
+         */}
+        {/* eslint-disable-next-line @next/next/no-css-tags */}
```
Comment (kasbah) [inline frontend/src/pages/[username]/[projectName]/IBOM.js:99]: Does `eslint-disbable` have to be a separate comment?

Comment (AbdulrhmnGhanem) [reply frontend/src/pages/[username]/[projectName]/IBOM.js:99]: Yes

frontend/src/components/Board/BuyParts/index.js
kasbah: > Comment line-length is too long now :D

frontend/src/components/NavBar.js
AbdulrhmnGhanem: This is wrapped with `Link`
https://github.com/kitspace/kitspace-v2/blob/df88c9bbde9cb32efc4c9ec71424679ba66fa38b/frontend/src/components/NavBar.js#L72-L74
`Link` passes a ref to its child, if the child is a functional commponent it has to be wrapped with forwardRef.

kasbah: But why? It seems kind of complicated for what it's doing. There must be a simpler way...

AbdulrhmnGhanem: I couldn't find any other solution. Maybe use `a` but we lose the features of the `Link`.

kasbah: Why use an `a` at all? Why not put the `Link` where the `a` is?

AbdulrhmnGhanem: This is the original code 
```jsx
<Link href="/" passHref>
  <Image
    width={160}
    height={46}
    objectFit="scale-down"
    className={styles.logoImg}
    src={logoSrc}
    alt="logo"
  />
</Link>
```
It causes this error in the browser, it's a runtime error, not eslint one
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```
If a `Link` is used in the `Logo` component
https://github.com/kitspace/kitspace-v2/blob/df88c9bbde9cb32efc4c9ec71424679ba66fa38b/frontend/src/components/NavBar.js#L28-L42
we will end up in the same situation just introduced some redundant wrapper.

kasbah: Why do we need `passHref`? I don't really understand why this is getting so complicated, all we want is an image with a link to `/`.

AbdulrhmnGhanem: Because using `Link` without `passHref` will work but not totally, for example, if you hover over the image it won't show the link of `/` at the bottom which works with the plain `a`. As I said before the simplest solution would be using
```jsx
<a href="/" >
  <Image
    width={160}
    height={46}
    objectFit="scale-down"
    className={styles.logoImg}
    src={logoSrc}
    alt="logo"
  />
</a>
```
and disable eslint rule for this line.

kasbah: Does `Link` never show the url at the bottom? If we use an `a` it will do a full refresh/navigation when you click on it which seems like a waste. If you can't find a way to clean it up please leave a comment with a reminder to clean it up down the line and an explanation of why it's like this.

AbdulrhmnGhanem: > Does Link never show the url at the bottom?

It works if the `passHref` prop is `true`.

AbdulrhmnGhanem:
> There's a problem with the eslint for next and babel [vercel/next.js#26127](https://github.com/vercel/next.js/issues/26127).

A workaround for this is to open the frontend directly in the editor, not the top-level repo.

PR #146, opened 2021-07-08
Add an API to processor for processing remote files

kasbah:
I'm doing some work that will allow rendering KiCad files on Github through our browser extension so I'm adding an API to the processor where the extension can send it files and get the result.


processor/src/remoteAPI.js
```diff
@@ -0,0 +1,109 @@
...
+        throw Error('Could not hash file')
+      }
+      const uploadFolder = path.join(remoteProcessInputDir, upload.md5)
+      const ext = path.extname(upload.name)
+      if (ext !== '.kicad_pcb') {
+        return res.status(422).send('Only accepting .kicad_pcb files')
+      }
+      const uploadPath = path.join(uploadFolder, upload.md5 + ext)
+      const outputDir = path.join(remoteProcessOutputDir, upload.md5)
+      await exec(`mkdir -p ${uploadFolder}`)
```
AbdulrhmnGhanem, line 62: Will these lines be vulnerable to [path traversal attacks](https://owasp.org/www-community/attacks/Path_Traversal)?

AbdulrhmnGhanem:
Don't forget updating allowed methods

https://github.com/kitspace/kitspace-v2/blob/ce9e772bed1e6257e7e00f41fe0e48e77bfcf4ed/nginx/kitspace.template#L93

processor/README.md
```diff
@@ -72,7 +72,35 @@ or
         - as above but prefixed with the multi-project name i.e. `[user]/[project]/[git-ref]/[multi-project-name]/zip-info.json` etc.
         - except for `kitspace-yaml.json` which is still `[user]/[project]/[git-ref]/kitspace-yaml.json`
 
+### Processing of remote files through API
 
+#### POST `/process-file`
+
+- Multi-part form data with an input name of "upload" and a filename that ends with `.kicad_pcb`
+- Must include a header `Authorization` with contents `Bearer ${KITSPACE_PROCESSOR_REMOTE_API_TOKEN}`
```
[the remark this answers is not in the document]
kasbah: It's a static "secret" set with an env variable. You have to know it when you make the request. I've set it via the Github environment secrets for our staging deploys. I don't think we will use this API at all on our staging servers for now. 

I will set it to something else for the initial deploy that will provide the API for the browser extension. It's stored unencrypted in the extension and sent along with requests.  There's no way to keep it properly secret and also use it in the extension so it is what it is. It just makes DDoSing the server slightly less convenient. I might put some restrictions on the request origin as well.

processor/src/remoteAPI.js
kasbah: I don't think so because there are no API user provided paths at all.

AbdulrhmnGhanem: where will `upload.md5` come from?

kasbah: It comes from the express-fileupload middleware.

AbdulrhmnGhanem: OK, I didn't know that :smile:

AbdulrhmnGhanem (approved):

PR #165, opened 2021-07-18
Fix checking collaborators causing 404

### LINKED ISSUE #164: 404 when checking for collaborators
Check if the user is part of the list of collaborators from `/[repo/collaborators` instead.

kasbah:
closes #164, seems like csrf is needed


kasbah:
It's a bit wrong. Still needs to be changed to actually check the user object returned.

AbdulrhmnGhanem:
The [failed test](https://dashboard.cypress.io/projects/d8hk55/runs/1511/test-results/f3b854d5-98ad-4f57-882c-b4c98994aee2) is the `404` happening on staging.

PR #182, opened 2021-08-06
Fix conflicting project names modal

### LINKED ISSUE #90: Conflicting project name modal buttons are a bit confusing
As discussed on the call, the buttons need to be hidden rather than disabled. The rename one should just be "OK" after you changed the name. The other one should say `Add file to ${projectName}` or something.

AbdulrhmnGhanem:
Fixes #90.
##
- The rename button is "OK" after changing the name.
- The upload to the same project is `Add file to ${projectName}`.
##
### Add files to the same project
image
##
### Changing the project name
image
##
### Changing the project name though the new name is also already used
image


kasbah:
That's definitely better. I think the color of the button, when adding files to an existing one, should be the warning colour (i.e. orange).

PR #184, opened 2021-08-06
Unify frontend and processor ibom

### LINKED ISSUE #144: Unify frontend and backend IBOM
https://github.com/kitspace/kitspace-v2/blob/c203c0363000b80e4c2ccee0ece5fc4e9c3e40e2/processor/src/tasks/processIBOM/run_ibom#L66-L67

This indicates there is a version mismatch of IBOM between the frontend and processor backend. We need to unify these.

AbdulrhmnGhanem:
Fixes #144.
##
**Note**: prune the staging server before testing this, because the IBOM/index.js file is read from a volume.

### DIFF
--- a/frontend/public/static/IBOM/index.js
+++ b/frontend/public/static/IBOM/index.js
@@ -1,4 +1,2996 @@
-/* DOM manipulation and misc code */
+/*
+  Split.js - v1.3.5
+  MIT License
+  https://github.com/nathancahill/Split.js
+*/
+!(function (e, t) {
+  'object' == typeof exports && 'undefined' != typeof module
+    ? (module.exports = t())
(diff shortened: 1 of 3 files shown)


kasbah:
If I'm reading this right you are simply updating the one in the frontend. What I meant by "unify" is to pull it in from a single git submodule in both frontend and processor.

AbdulrhmnGhanem:
> What I meant by "unify" is to pull it in from a single git submodule in both frontend and processor.

Added a script to do so in af9ab3e.

AbdulrhmnGhanem:
> What I meant by "unify" is to pull it in from a single git submodule in both frontend and processor.

Added a script to do so in af9ab3e.

PR #202, opened 2021-09-10
Move readme processing to the `processor`

### LINKED ISSUE #128: Sort out readme processing
Right now it's done in Gitea but needs some post-processing to sort out the links. We should decide where we want to do the processing. Ideally we wouldn't have a cheerio dependency in the frontend.

https://github.com/kitspace/kitspace-v2/blob/e4e9ab98ed31d8bc61246ce5b69570dec681323c/frontend/src/utils/index.js#L37-L88

AbdulrhmnGhanem:
Closes #128


docker-compose.override.yml
```diff
@@ -42,6 +42,7 @@ services:
       - '5000:5000'
     command: /bin/bash --verbose -c 'yarn install && yarn dev'
     volumes:
+      - /etc/hosts:/etc/hosts
```
kasbah, line 45:
```suggestion
      - /etc/hosts:/etc/hosts:ro
 ```

File: processor/src/tasks/processReadme.js [hunk 2]
```diff
@@ -0,0 +1,127 @@
...
+
+  if (readmeFiles.length >= 1) {
+    return readmeFiles[0]
+  } else {
+    return null
+  }
+}
+
+async function renderMarkdown(rawMarkdown) {
+  const giteaMarkdownEndpoint = `${GITEA_URL}/api/v1/markdown/raw`
```
kasbah, line 86: Instead of accessing it throught the external domain we can go through docker networking, i.e. 

```
  const giteaMarkdownEndpoint = `http://gitea:3000/api/v1/markdown/raw`
```

processor/src/tasks/processReadme.js
```diff
@@ -0,0 +1,127 @@
...
+}
+
+async function _processReadme(
+  events,
+  inputDir,
+  kitspaceYaml,
+  outputDir,
+  projectFullname,
+) {
+  const readmePath = path.join(outputDir, 'readme.json')
```
kasbah, line 38: Why `.json`?

AbdulrhmnGhanem: The file extension doesn't really matter. If it were `.html` the only difference in the frontend `res.text()` would be used instead of `res.json()`

kasbah: Except it just obfuscates the actual content and wraps in JSON for no reason. Please make this `.html`.

processor/src/tasks/processReadme.js
AbdulrhmnGhanem: Resolved in c0f4757.

processor/src/tasks/processReadme.js
AbdulrhmnGhanem: Resolved in a2a5dc0.

PR #205, opened 2021-09-14
Cache static file with nginx

AbdulrhmnGhanem:


AbdulrhmnGhanem:
This is no longer irrelevant as we have the assets in S3 now. We can get back to this when we proxy the s3 bucker through our sever.

PR #227, opened 2021-11-22
Improve authentication redirect

### LINKED ISSUE #190: Auth checking and redirect needs to be done on next pages level
as discussed on call

AbdulrhmnGhanem:
Fixes #190
##
- The newly added `_middleware.js` only runs for server-side requests,
- `next-composition` works for both server-side and client-side requests.
- Before the redirection for pages that required sign in, e.g., `projects/new`, and sign out, e.g., `/login` was done by checking the `AuthContext` in the client-side.
- Now it is done with getInitialProps, the logic is unified in a single unit (`utils/authHandlers`) with the help of `next-composition`.


AbdulrhmnGhanem:
The layout SSR bug fixed in


|before|after|
|---|---|
image|image|

frontend/src/components/Page.js
```diff
@@ -62,40 +20,39 @@ const Container = ({ contentFullSize, children }) => (
...
+    const browserPath = asPath.split('?')[0]
+    const doesTheBrowserURLMismatchPathname = browserPath !== pathname
+    if (
+      doesTheBrowserURLMismatchPathname &&
+      pathname !== '/login' &&
+      browserPath !== '/'
+    ) {
+      replace(pathname, null, { shallow: true })
+    }
+  }, [asPath, pathname, replace])
```
kasbah, line 36: Why is this needed? Why can they still be mismatched?

AbdulrhmnGhanem: Added docs in 08e190b.

PR #234, opened 2022-01-09
Fix premature processing for gitea repos

### LINKED ISSUE #100: The processor stuck at `addDir` when processing repos
Sometimes after migrating the first repo to gitea, the processor gets stuck at the `addDir` step. restarting the container fixes the problem.

AbdulrhmnGhanem:
Fixes #100
## 
- When migrating a repo [1]: Gitea creates an empty repo then the migration
  starts.
- When uploading user files [2]: we create an empty repo,
  upload the files, finally commit the files to the repo.
- When the processor gets notified by a file-system change when the
  repository is created in both [1], and [2] the repo is still empty and
  the `git checkout` command fails.
- The fix:
    - only process repos if it isn't empty,
    - only process synced repos when the migration task is done.
 ##
* The processor check gitea database for the migration state of the repo.
* The maximum time for retries can be configured in the evn via `MAXIMUM_REPO_MIGRATION_TIME`.
##
* `e2e` doesn't have any flaky tests now :tada:


processor/src/processingManager.js
```diff
@@ -0,0 +1,233 @@
...
+      .backoff()
+  })
+}
+
+const ProcessingManager = {
+  client,
+  isRepoReadyForProcessing,
+}
+
+module.exports = { ProcessingManager }
```
kasbah, line 233: We need a better name for all of this. Name it after what it's doing, it's definitely not "managing" the processing.

processor/src/processingManager.js
```diff
@@ -0,0 +1,233 @@
...
+
+    const onFail = reject =>
+      reject(
+        new Error(
+          `Repo: ${ownerName}/${repoName} not found after ${MaximumRetries} trials.`,
+        ),
+      )
+
+    return asyncBackoff(onBackoff, onFail, MAXIMUM_NUM_OF_RETRIES)
+  }
```
kasbah, line 132: Can this function be defined outside the top level function?

processor/src/processingManager.js
```diff
@@ -0,0 +1,233 @@
...
+
+    const onFail = reject =>
+      reject(
+        new Error(
+          `Repo: ${repoId} migration task not found after ${MaximumRetries} trials.`,
+        ),
+      )
+
+    return asyncBackoff(onBackoff, onFail, MAXIMUM_NUM_OF_RETRIES)
+  }
```
kasbah, line 181: Here too: can this function be defined outside the top level function?

processor/src/processingManager.js
kasbah: I'm wondering about the backoff as well: can we use a Postgres/SQL lib that would take care of the lowere level details?

processor/src/app.js
```diff
@@ -2,15 +2,15 @@ const express = require('express')
 
 const { createProjectsAPI } = require('./projectAPI')
 const { createRemoteAPI } = require('./remoteAPI')
-const { ProcessingManager } = require('./processingManager')
+const { isRepoReadyForProcessing } = require('./isReadyForProcessing')
```
kasbah, line 5: I generally try and follow this rule: `is..` is a boolean variable, never a function. So I would suggeest: `checkIsRepoReady` and call the file `checkIsRepoReady`.

processor/src/processingManager.js
AbdulrhmnGhanem: Renamed it to `isReadyForPorcessing`.

processor/test/integration/test_projects.js
```diff
@@ -31,7 +31,7 @@ describe('projects API', () => {
   beforeEach(async () => {
     await exec(`mkdir -p ${tmpDir}`)
     await exec(`mkdir -p ${repoDir}`)
-    this.app = createApp(repoDir, processingManagerStub)
+    this.app = createApp(repoDir, isRepoReadyForProcessingStub)
     this.supertest = supertest(this.app)
   })
```
kasbah, line 37: That stub is really ugly, can you not let it be `undefined` and check it down the line?

processor/src/watcher.js
```diff
@@ -24,7 +33,7 @@ function watch(events, repoDir = '/repositories') {
     // additionally we ignore any invocations that happen while it's already running
     // to prevent it from trying to overwrite files that are already being written to
     const debouncedProcessRepo = debounce(async () => {
-      if (!running[gitDir]) {
+      if ((await isRepoReadyForProcessing(gitDir)) && !running[gitDir]) {
```
kasbah, line 36:
```suggestion
const isReady =  isRepoReadyForProcessing == null || (await isRepoReadyForProcessing(gitDir))
if (!running[gitDir] && isReady) {
```

processor/src/watcher.js
```diff
@@ -13,7 +14,15 @@ const processKicadPCB = require('./tasks/processKicadPCB')
 const processReadme = require('./tasks/processReadme')
 
 const running = {}
-function watch(events, repoDir = '/repositories') {
+
+/**
+ *
+ * @param {*} events
+ * @param {string=} repoDir
+ * @param {checkIsRepoReady} checkIsRepoReady
```
kasbah, line 22:
```suggestion
 * @param {function=} checkIsRepoReady
```

PR #240, opened 2022-01-24
Fix: warnings in `/login`

AbdulrhmnGhanem:


kasbah:
Can you explain what's going on here please.

frontend/src/pages/login/index.js
```diff
@@ -40,7 +40,6 @@ const Login = () => {
         <Grid.Column>
           <Tab
             activeIndex={openPane}
-            defaultActiveIndex={0}
```
AbdulrhmnGhanem, line 43: `defaultActiveIndex` shouldn't be used when `activeIndex` is used.

frontend/src/pages/login/index.js
```diff
@@ -282,7 +281,6 @@ const SignUpForm = ({ openLoginPane }) => {
             onClick={submit}
           />
           <Form.Field
-            basic
             as="a"
```
AbdulrhmnGhanem, line 284: `Form.Field` doesn't have a `basic` prop.

PR #241, opened 2022-01-24
Fix multi projects redirect

AbdulrhmnGhanem:
- redirect to subproject on clicking the card on multi-project page
   - For multi-project, visiting `/[username]/[projectName]` displays a
    grid of subprojects. Clicking on any of the cards didn't redirect to
    the correct subproject page.
    - Update `multi-project.spec` to catch this bug.
- clicking on the IBOM title for multi-project doesn't redirect to the project page.
  - Added tests for this case for normal and multi-projects

### DIFF
--- a/frontend/cypress/integration/IBOM.spec.js
+++ b/frontend/cypress/integration/IBOM.spec.js
@@ -1,0 +1,101 @@
+import faker from 'faker'
+
+import { getFakeUsername } from '../support/getFakeUsername'
+
+describe('IBOM page', () => {
+  before(() => {
+    /*
+     * The purpose of this isn't actually visiting the homepage.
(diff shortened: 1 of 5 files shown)


frontend/src/components/SharedProjectPage/elements.js
```diff
@@ -1,5 +1,5 @@
 import React, { useEffect, useState, useContext } from 'react'
-import { bool, objectOf, string, object, number, shape } from 'prop-types'
+import { bool, objectOf, string, number, shape } from 'prop-types'
 import { useRouter } from 'next/router'
```
kasbah, line 3: Can you sort out the merge conflict with these prop-types please? I think `object` is used below so I'm confused.

PR #265, opened 2022-02-01
Switch to swc in frontend

kasbah:


kasbah:
@AbdulrhmnGhanem please check https://review.staging.kitspace.dev to see if there could be any issues with this that I haven't noticed.

AbdulrhmnGhanem:
> Oh, I'm not sure what to do about @welldone-software/why-did-you-render. Do we need it? Will it work with swc?

I don't think it will work, because it patches react in the babel config. For now, you can remove it. If we ever need it again will look for a solution then.

kasbah:
Hmm, well, if it's useful and there's no replacement then I am not so sure. Switching to `swc` doesn't seem to get us much. Build times seemed fine before it. I'm mainly proposing it because the babel setup with styled-components seems really buggy/brittle. I can't seem to upgrade to a newer styled-components.

kasbah:
The lodash transform is another one, we should check the bundle size if we want to make this switch.

kasbah:
Closing in favor of #265

PR #268, opened 2022-02-02
Fix blinking of logo component on Firefox

### LINKED ISSUE #266: Logo re-rendering for no reason
When you navigate around the logo blinks in and out. No idea why, it should just be staying in place.

kasbah:
I think this fixes #266. I'm still confused about this one though, about the ref passing etc. not sure what it was for at all. I never saw any warnings about it. Maybe it was fixed?

### DIFF
--- a/frontend/cypress/integration/IBOM.visual.spec.js
+++ b/frontend/cypress/integration/IBOM.visual.spec.js
@@ -59,15 +59,6 @@ describe('Regression test for IBOM ', () => {
   )
 
   it(
-    'should match the existing snapshot for the IBOM (firefox)',
-    { browser: 'firefox' },
-    () => {
-      // Compare/save the snapshot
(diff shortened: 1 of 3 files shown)


kasbah:
Ah found the docs about it: https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-functional-component (that should have been in the comment!)

Anyway, it doesn't seem to fix #266

AbdulrhmnGhanem:
The navbar height has increased, is it intentional?

kasbah:
Not really, but by a pixel or something? Seems fine... nav bar has some layout shift issues anyway so will need to take a close look at the heights of things at some point.

kasbah:
@AbdulrhmnGhanem can you update the snapshots for now please? (Though we need a better way of doing this as mentioned in #290.)

PR #271, opened 2022-02-02
Stop removing buttons from nav bar based on location

kasbah:
It actually seems annoying, lol. Let's just keep the nav bar the same all the time.


kasbah:
Merged in with #295

Issue #272, opened 2022-02-02
Is the whole page being remounted on transitions?


kasbah:
I'm trying to figure out what's going on with the page transitions and how we can fix it. The experience I'd like to get back to is what you get with a regular next app. 

Here is [a demo repo](https://github.com/kasbah/next-app-test). You can clone that, run the dev server, and click between "go to page two" and "go to page one". That's the kind of page transition we want, what's happening with ours? We get layout shift, flash of a loader, and issues like #268. Is everything being dismounted and mounted again unnecessarily?

kasbah:
Alright, let me know if you run out of ideas and I'll continue my investigation. One thing I did do was run the [almost vanilla create-next-app behind the `/__kitspace` gitea hack to check if it's that](https://github.com/kitspace/kitspace-v2/tree/kaspar/debug-page-transitions/frontend) (it's not).

AbdulrhmnGhanem:
I am trying this [blog](https://adamwathan.me/2019/10/17/persistent-layout-patterns-in-nextjs/) suggestions.

kasbah:
Those blog suggestions seem to be complicating things even more. I feel like we had this and then lost it, hence we have it when go back to vanilla create-next-app. (Even if create-next-app is still doing more re-rendering than needed according to that blog.) I suspect it's something to do with the auth and the use of `useEffect` which we never fully straightened out.

kasbah:
I think most of this is solved with #280, there's still some "blinking" I can see on Firefox but I should probably make a more specific issue.

PR #280, opened 2022-02-02
Fix menu items not using next/link

kasbah:
This actually solves most of what I was complaining about with https://github.com/kitspace/kitspace-v2/issues/272 (I was mostly clicking on these menu items). There seems to be a bit of "blinking" still. I think that's down to the way stuff is painted before auth is checked. The brief flash of the spinner on the project pages is another matter.


AbdulrhmnGhanem:
You beat to reporting it :cry:

PR #284, opened 2022-02-04
Remove `babel-plugin-lodash`

AbdulrhmnGhanem:
Note this is against the `frontend-upgrade-deps-2` not `master`.


kasbah:
I don't think I want to go ahead with #265

kasbah:
Not sure what "draft" means in this context. Am I ok to merge this into #265 (and probably abandon it there)?

PR #291, opened 2022-02-05
Turn on gzip compression in nginx

kasbah:


nginx/nginx.conf
```diff
@@ -27,7 +27,7 @@ http {
 
     keepalive_timeout  65;
 
-    #gzip  on;
+    gzip  on;
```
AbdulrhmnGhanem, line 30: Nginx by default compresses html only.
```suggestion
    gzip  on;
    gzip_types text/css application/javascript application/json;
```

kasbah: Looking at the network tab: seems to be compressing js, css  and html by default and missing fonts (woff2), svg and json actually.

AbdulrhmnGhanem: Add them just in case.

> By default, NGINX compresses responses only with MIME type text/html. To compress responses with other MIME types, include the [gzip_types](https://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_types) directive and list the additional types.
— https://docs.nginx.com/nginx/admin-guide/web-server/compression/

AbdulrhmnGhanem: > seems to be compressing js, css and html by default 

I think this is [`Next` compression](https://nextjs.org/docs/api-reference/next.config.js/compression) not Ngnix.

AbdulrhmnGhanem (approved):

kasbah:
Investigating why woff2 compression doesn't seem to be happening.

PR #295, opened 2022-02-08
Streamline authentication

kasbah:
- Removes the hacky /__kitspace routing via gitea
- Uses client-side `push` for redirects as much as it can
- Uses `getServerSideProps` redirects for "requires sign in" redirects

Gitea changes are on https://github.com/kitspace/gitea/pull/23.

(I also merged #271 into this because I updated the tests to make everything pass and it's a hassle to separate them out again)


frontend/cypress/integration/signInForm.spec.js
```diff
@@ -44,43 +44,18 @@ describe('Log in form submission', () => {
...
-  })
-
-  it('should display error message on submitting form for prohibited user', () => {
-    cy.stubSignInReq(false, {
-      error: 'Prohibited',
-      message: 'Prohibited login.',
-    })
-    cy.signIn('prohibitedUser', password)
-    cy.get('.negative').should('include.text', 'Prohibited login.')
-  })
```
kasbah, line 85: Removing these because all the `stub` things had weird errors on my test runs that I had no idea how to resolve.

frontend/src/contexts/AuthContext.js
```diff
@@ -1,47 +1,47 @@
...
+  const { push } = useRouter()
 
   useEffect(() => {
-    const session = window?.session
-
-    if (session?.user) {
-      authorize(session.user, session._csrf)
-    } else {
-      deAuthorize()
+    if (typeof window !== 'undefined') {
```
AbdulrhmnGhanem, line 23: What is the purpose of this check? Hooks don't run on the server-side.

frontend/src/pages/_app.js
```diff
@@ -34,17 +34,27 @@ import './_app.scss'
...
+  isStaticFallback,
+}) {
+  const setSession = serverSideSession ? (
     <script
+      // using dangerouslySetInnerHTML to avoid browser parsing errors
+      // (could be that these only happen in development mode)
       // eslint-disable-next-line react/no-danger
       dangerouslySetInnerHTML={{
-        __html: `window.session = ${JSON.stringify(session)};`,
+        __html: `window.session = ${JSON.stringify(serverSideSession)};`,
```
AbdulrhmnGhanem, line 50: Isn't this redundant? Previously the values of `user` and `csrf` were read from `window.session` now it's passed directly to the context.

frontend/src/pages/_app.js
```diff
@@ -55,7 +65,10 @@ function KitspaceApp({ Component, pageProps, session, isStaticFallback }) {
     isStaticFallback = isStaticFallback || window.isStaticFallback
   }
   return (
-    <AuthProvider>
+    <AuthProvider
+      initialCsrf={serverSideSession?.csrf || clientSideSession?.csrf}
+      initialUser={serverSideSession?.user || clientSideSession?.user}
+    >
```
AbdulrhmnGhanem, line 71: From the `AuthProvider` perspective, is there a difference between `serverSideSession`, and `clientSideSession`? I think `KitspaceApp` should get a single `session` Prop.

frontend/cypress/integration/authRedirects.spec.js
```diff
@@ -138,22 +138,20 @@ describe('It validates redirects after sing up', () => {
...
 })
 
-describe('It validates redirect after logout', () => {
+describe('Redirect after logout', () => {
   it('should redirect to login page', () => {
     const username = getFakeUsername()
     const email = faker.unique(faker.internet.email)
     const password = '123456'
     cy.createUser(username, email, password)
     cy.visit('/login')
```
AbdulrhmnGhanem, line 154: Shouldn't this get deleted too?

```suggestion

```

frontend/cypress/integration/authRedirects.spec.js
```diff
@@ -180,14 +178,19 @@ describe('It validates `Add Project` behavior', () => {
...
     // Clicking `Add Project` redirects to the login page.
     // and adds redirect query to `/projects/new/`
     cy.get('#add_project').click()
-    cy.url().should('eq', 'http://kitspace.test:3000/login?redirect=/projects/new')
+
+    const username = getFakeUsername()
+    const email = faker.unique(faker.internet.email)
+    const password = '123456'
+    cy.createUser(username, email, password)
+    cy.url().should('eq', 'http://kitspace.test:3000/projects/new')
```
AbdulrhmnGhanem, line 193: Checking redirecting to `http://kitspace.test:3000/login?redirect=/projects/new` is deliberate. Because this one checks clicking on the `Add project` button. There's another test that checks redirecting to the page specified in the query param.
https://github.com/kitspace/kitspace-v2/blob/6fddd565bd2eecd7db2a2a12a15c9253362a41be/frontend/cypress/integration/authRedirects.spec.js#L78

frontend/cypress/integration/signInForm.spec.js
kasbah: You shouldn't be rolling this back! The other reason to remove this by the way is that we shouldn't be baking the http API into the e2e tests.

kasbah: Also, while it may not be showing up on the CI these stubs were definitely causing problems running e2e locally on my machine.

frontend/src/contexts/AuthContext.js
kasbah: Paranoia

frontend/src/pages/_app.js
kasbah: From the `AuthProvider` perspective, no. But from the `KitspaceApp` perspective, they need to be handled very differently. That's why I gave them clearer names.

frontend/src/pages/_app.js
kasbah: I thought so too at first and tried to remove it, but it's still needed. It makes sense when you think about it. The only way to for the client-side to get the session. when you use `getInitialProps`, is for the server to embed it into the page. 

I can see a cleaner solution by using `getServerSideProps` on every page to get the session from gitea, rather than us a custom _app.js (we can make something re-usable like we did for `withRequireSignIn)`, but let's leave that for another PR.

frontend/cypress/integration/authRedirects.spec.js
kasbah: I'm not sure we should have e2e tests specific to the query parameter. With all of these tests I'm thinking about "what does the user care about in the moment". The user doesn't care if there's a `redirect` parameter or not. They care that they are redirected to the right place. Maybe we will change the query parameter at some point and use session storage to redirect. Should the test fail because we changed that?

I even think we should remove all the `cy.url().should('eq')` checks and replace them with `cy.get` for page-specific elements. (Of course, once we go public, we'd like to keep URLs consistent and can add checks for them then, but even then I'm not sure we should be checking a query parameter for the login redirect.)

frontend/src/pages/_app.js
AbdulrhmnGhanem: We can keep it for now and get rid of it with `window.session`.

AbdulrhmnGhanem (approved):

frontend/cypress/integration/authRedirects.spec.js
AbdulrhmnGhanem: I try to test my assumptions about how it should behave, not just user interaction it makes changing any assumption doesn't go unnoticed.

kasbah: I'm not quite sure what you are saying, sorry. By "this makes changing.."  do you mean "this means that changing..."? What I'm just saying is the tests should emulate user behavior as much as possible and not rely on implementation details. You can still check your assumptions while not doing that.

AbdulrhmnGhanem: > I'm not quite sure what you are saying

> I try to test my assumptions about how it should behave

Anyways, we can test the implementation in unit tests.

AbdulrhmnGhanem (approved):

PR #298, opened 2022-02-08
Re-enable schematic processing

kasbah:
Re-enable schematic.svg output in the processor now that we can set concurrency limits.

### DIFF
--- a/processor/README.md
+++ b/processor/README.md
@@ -64,6 +64,7 @@ or
         - `images/top-meta.png`
         - `images/top-with-background.png`
         - `images/layout.svg`
+        - `images/schematic.svg`
         - `1-click-BOM.tsv`
         - `bom-info.json`
         - `interactive_bom.json`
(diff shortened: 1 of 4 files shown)


kasbah:
Hmmm.. seems to be still causing some issues.

AbdulrhmnGhanem (commented):
Why does this error happen when migrating a repo? Does `plot_kicad_sch_docker` get triggered unnecessarily?

```console
processor_1  | {
processor_1  |   stdout: '',
processor_1  |   stderr: '\n' +
processor_1  |     'input_sch="$1"\n' +
processor_1  |     'output_dir="$2"\n' +
processor_1  |     '\n' +
processor_1  |     '# how volumes work here is a bit confusing but this docker is essentially the\n' +
processor_1  |     '# outer-most docker since it shares the docker.sock. we mount the\n' +
processor_1  |     '# processor-data volume in the same place meaning all our paths match up as\n' +
processor_1  |     '# long as they are under /data\n' +
processor_1  |     'docker run --rm \\\n' +
processor_1  |     '    --volume="${COMPOSE_PROJECT_NAME}_processor-data:/data:rw" \\\n' +
processor_1  |     '    setsoft/kicad_auto /bin/bash -c \\\n' +
processor_1  |     '    "eeschema_do export -f svg \'${input_sch}\' \'${output_dir}\'"\n' +
processor_1  |     'WARNING:Missing KiCad main config file /root/.config/kicad/kicad_common (eeschema_do - misc.py:176)\n'
processor_1  | }
```

kasbah:
I'll look into it. In general when I migrated to BullMQ I got rid of some checks that only allowed one task to write the same file at a time. I kind of forgot about it because it wasn't causing any issues. Thanks for the reminder.

PR #300, opened 2022-02-09
Debug `Install gitea` step in the CI

AbdulrhmnGhanem:
It's failing without applying any change it


AbdulrhmnGhanem:
Actually it was failing for a reason.

PR #301, opened 2022-02-10
Update gitea

kasbah:


AbdulrhmnGhanem:
Creating the admin before running the tests isn't needed anymore so removed it in 1b30930.

PR #309, opened 2022-02-13
Convert processor to typescript

kasbah:


processor/src/server.ts
```diff
@@ -0,0 +1,12 @@
+import * as log from 'loglevel'
+import { createApp } from './app'
+import { checkIsRepoReady } from './checkIsRepoReady'
+
+log.setDefaultLevel((process.env.LOG_LEVEL as log.LogLevelDesc) || log.levels.INFO)
```
AbdulrhmnGhanem, line 5: I think type assertion should be in `env.ts`.

AbdulrhmnGhanem (approved):

PR #324, opened 2022-04-07
Fix bad substitution error for nginx

kasbah:
I think something about how docker-compose escapes the envs changed. I recently updated to `Docker Compose version 2.3.3` and was getting `sed` errors. Can you check if changing this in your .env breaks your nginx @AbdulrhmnGhanem.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,4 +22,4 @@ MAX_FILE_SIZE=6M
 # The maximum time the migrating repo to gitea can take (in ms),
 # after which the processor will consider the migration a failure
 MAXIMUM_REPO_MIGRATION_TIME=300000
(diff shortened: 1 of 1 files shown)


AbdulrhmnGhanem:
`docker-compose docker-compose build --no-cache nginx && docker-compose up` is working as usual.

kasbah:
Both with and without this change? :thinking:

AbdulrhmnGhanem:
Yes, I ran the command on both master and this branch.

kasbah:
I'm only changing the .env.example here though, .env isn't in git of course and is the one you have to change.

AbdulrhmnGhanem:
I missed it the first time I tried but updated it and ran the command again on the master and this branch, it works as expected, and visiting `/robots.txt` returns the same response.

PR #330, opened 2022-04-12
[TEST PR] limit e2e matrix across all browsers to `push` events on `master`

AbdulrhmnGhanem:


AbdulrhmnGhanem:
It turns out that even if I change the `pull_request_target` in this branch it still runs the workflow in the master branch so this test is useless we have to merge it into master and see!

PR #351, opened 2022-05-18
Fix processor running out of memory

kasbah:
Fixes the memory issue of the processor we encountered testing out the d20 project. 

- Uses `jobId` to not start jobs that are already in progress
- Limits worker concurrency to 1 (i.e. no concurrency of an individual worker type)
- Removes concurrency of `inkscape` executions which take up a lot of memory

d20 was now imported without issue on https://review.staging.kitspace.dev/ 

Additionally it now removes/`obliterates` all queues when stopped (which clears out the redis data). This was needed for the integration tests, probably should make use of this when exiting during normal operation as well.

### DIFF
--- a/processor/src/tasks/processGerbers/index.ts
+++ b/processor/src/tasks/processGerbers/index.ts
@@ -113,21 +113,17 @@ export default async function processGerbers(
         job.updateProgress({ status: 'failed', file: topSvgPath, error }),
       )
 
-    promises.push(
-      generateTopPng(topSvgPath, stackup, topPngPath)
-        .then(() => job.updateProgress({ status: 'done', file: topPngPath }))
-        .catch(error =>
-          job.updateProgress({ status: 'failed', file: topPngPath, error }),
-        ),
-    )
+    await generateTopPng(topSvgPath, stackup, topPngPath)
+      .then(() => job.updateProgress({ status: 'done', file: topPngPath }))
+      .catch(error =>
+        job.updateProgress({ status: 'failed', file: topPngPath, error }),
+      )
 
-    promises.push(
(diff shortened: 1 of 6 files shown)


processor/src/tasks/processGerbers/index.ts
```diff
@@ -113,35 +113,29 @@ export default async function processGerbers(
...
-        .then(() => job.updateProgress({ status: 'done', file: topWithBgndPath }))
-        .catch(error =>
-          job.updateProgress({ status: 'failed', file: topWithBgndPath, error }),
-        ),
-    )
+    await generateTopWithBgnd(topMetaPngPath, topWithBgndPath)
+      .then(() => job.updateProgress({ status: 'done', file: topWithBgndPath }))
+      .catch(error =>
+        job.updateProgress({ status: 'failed', file: topWithBgndPath, error }),
+      )
```
AbdulrhmnGhanem, line 138: > Removes concurrency of inkscape executions which take up a lot of memory

I guess this is the code for this change, a comment would be helpful to say these are done sequentially or even better group them into a task.

AbdulrhmnGhanem (approved):

processor/src/tasks/processGerbers/index.ts
```diff
@@ -113,35 +113,29 @@ export default async function processGerbers(
         job.updateProgress({ status: 'failed', file: topSvgPath, error }),
       )
```
kasbah, line 115:
```suggestion

// The generate*Png tasks shouldn't run concurrently so we `await` them in
// sequence.  They all invoke Inkscape and can use a lot of RAM so we run them
// one at a time even though they don't depend on each other.
```

@AbdulrhmnGhanem

PR #325, opened 2022-04-07
Use meilisearch

kasbah:

PR #353, opened 2022-05-21
Refactor gitea DB access in processor

kasbah:
I wanted the case correct names from the DB for adding to search in #325 and the DB access was doing my head in. I switched it to [porsager/postgres](https://github.com/porsager/postgres) and learned a bit about postgres replication/subscription. Not sure that subscription is really the right approach either (I asked a [question](https://github.com/porsager/postgres/discussions/372) about this) but it beats handling the polling backoff logic ourselves.


docker-compose.override.yml
```diff
@@ -27,6 +27,10 @@ services:
     volumes:
       - ./gitea:/go/src/code.gitea.io/gitea
 
+  postgres:
+    ports:
+      - '5432:5432'
```
AbdulrhmnGhanem, line 32: Why do we need this?

kasbah: It's handy for using `psql` while developing. We expose a bunch of other ports for development by the way.

AbdulrhmnGhanem (approved):
This is a hack, but a really neat one :+1:

docker-compose.override.yml
AbdulrhmnGhanem: OK!

FYI, I use this line `docker exec -it  <container-name/id> psql -U <postgres-username>` to get a postgres REPL inside a _running_ container.

PR #357, opened 2022-05-23
e2e: add version control info when running inside docker

AbdulrhmnGhanem:
Fixes ##355
##
I believe this will fix the badge too!

image

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -286,6 +286,11 @@ jobs:
           GITEA_DEPLOY_IMAGE_TAG: ':${{ needs.get_version.outputs.gitea_sha_short }}'
           browser: ${{ matrix.browser }}
           group: ${{ matrix.browser }}
+          COMMIT_INFO_MESSAGE: ${{ github.event.head_commit.message }}
+          COMMIT_INFO_EMAIL: ${{ github.event.head_commit.author.email }}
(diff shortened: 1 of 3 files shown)


AbdulrhmnGhanem:
The failure is due to the usual `push` vs `pull_request_target`.

.github/workflows/docker.yml
```diff
@@ -270,7 +270,14 @@ jobs:
         run: |
           scripts/install_gitea.sh
 
-      - name: pre e2e
+      - name: Generate build ID
+        id: uuid
```
kasbah, line 274: UUID has a very precise meaning and this doesn't seem to be that. https://en.wikipedia.org/wiki/Universally_unique_identifier

.github/workflows/docker.yml
```diff
@@ -286,6 +293,13 @@ jobs:
           GITEA_DEPLOY_IMAGE_TAG: ':${{ needs.get_version.outputs.gitea_sha_short }}'
           browser: ${{ matrix.browser }}
           group: ${{ matrix.browser }}
+          BUILD_ID: ${{ steps.uuid.outputs.value }}
```
kasbah, line 296:
```suggestion
          BUILD_ID: ${{ github.run_id }}-${{ github.run_attempt }}
```

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

kasbah (approved):

PR #363, opened 2022-05-23
Fix images for readmes in folders

kasbah:


kasbah:
@AbdulrhmnGhanem can you take another quick look please. I've enabled typescript in the frontend because I created a new file (meili.ts) and would like to only create new ts/tsx files to incrementally adopt it.

PR #368, opened 2022-05-24
Stop using ts-node in production in processor

kasbah:


AbdulrhmnGhanem:
#417 replaces this PR.

PR #370, opened 2022-05-24
Set "production" when installing deps in Dockerfile

kasbah:


AbdulrhmnGhanem:
I think I understand it now!
1. Install all packages
2. Build the production version of the app, generating `.next`
3. Install the prod packages inside the `/deps` dir
4. Copy `.next` (from 2), packages (from 3), `server.js`, and the static assets  into a new container

Is this right?

frontend/Dockerfile
```diff
@@ -1,15 +1,29 @@
...
-RUN mkdir /app
+FROM node:16-alpine AS base
 WORKDIR /app
 
-COPY package.json yarn.lock ./
-
-RUN yarn --frozen-lockfile
+FROM base AS git-installed
+## `git` is needed to install `1-click-bom-minimal` in `package.json`.
+RUN apk --no-cache add git
```
kasbah, line 7:
```suggestion

# The build stage installs the development dependencies, runs the build and
# then moves to another folder to install the production/non-dev dependencies.
# The build folder (.next) and the production dependencies (/deps/node_modules)
# are then copied out of this stage into the next one resulting in a more
# minimal image. 
```

kasbah:
Oh, nicely spotted. I think I could add add the `/public` folder to the base docker layer to fix it but I'm not sure if the Nginx serving of `/public` is getting us anything. Next seems to be replacing the src with `/_next/image?...` .

AbdulrhmnGhanem:
The logo is broken

image

kasbah:
I did the old radically unnecessary `docker-compose down -v && docker system prune --force --all --volumes && docker-compose up` and the logo and other assets still show up for me.

Maybe you have something cached somewhere? Or maybe it's down to docker versions?

```
$ docker-compose --version
Docker Compose version 2.5.1
$ docker --version
Docker version 20.10.16, build aa7e414fdc
```

AbdulrhmnGhanem:
It works now!

AbdulrhmnGhanem (approved):

kasbah:
I think the nginx rebuild was the one that was required.

PR #400, opened 2022-06-23
Encode subproject names

### LINKED ISSUE #359: `Open Visual Stimulator` Multi paths
https://github.com/eulerlab/open-visual-stimulator

The project is imported successfully 

Though clicking on any project card doesn't redirect to the multi-project; its links are broken. The mutlti-project names includes `/` which should be escaped!
```yaml
# kitsapce.yaml
multi:
    boards/mosfet_circuit:
      ...
    boards/lcr_add-on:
      ...
    boards/ssr_board:
      ...
```
screencapture-abdo-dev-staging-kitspace-dev-tester-open-visual-stimulator-2022-05-23-22_37_36

AbdulrhmnGhanem:
Fixes #359.

### DIFF
--- a/frontend/src/components/ProjectCard/index.js
+++ b/frontend/src/components/ProjectCard/index.js
@@ -10,7 +10,7 @@ import styles from './index.module.scss'
 const ProjectCard = ({ name, summary, ownerName, multiParentName }) => {
   const isMultiProject = multiParentName != null
   const repoName = isMultiProject ? multiParentName : name
-  const decodedName = decodeURIComponent(name)
+  const encodedName = encodeURIComponent(name)
   const { src, isLoading, isError } = useThumbnail(
     `${ownerName}/${repoName}`,
     multiParentName ? name : null,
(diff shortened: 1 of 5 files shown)


AbdulrhmnGhanem:
If `1-boards%2Fssr_board` or `1-boards/ssr_board` is used for `id` value in `meilisearch` document, the document doesn't get added to meilisearch.
https://github.com/kitspace/kitspace-v2/blob/17ed0a79d04413cf8d03371841e17e86deced977/processor/src/tasks/addToSearch.ts#L57-L69

https://docs.meilisearch.com/learn/core_concepts/primary_key.html#primary-field
What should I use instead?

kasbah:
Just remove or replace any characters that are invalid.

kasbah:
Thinking about this a bit further I think we should disallow characters in the name that would need URI escaping. We should have the same restrictions for sub-project names as we do for projects. Let's reject kitspace.yaml with these characters. We'll need to update the kitspace.yaml for Open Visual Stimulator.

AbdulrhmnGhanem:
So should I replace this PR with a validation step?

Issue #376, opened 2022-05-27
`4_LED_Channels` processing fails


AbdulrhmnGhanem:
https://kitspace.org/boards/github.com/badenlab/led-zappelin/pcb/potentiometer_mounts/4_led_channels/

AbdulrhmnGhanem:
We have three problems here:
1. sub-project names have slashes, already solved in #402.
2. The same race condition that happens in #392.
3. IBOM processing fails with the following error
```console
kitspace-processor-1  | failed tester/led-zappelin/dc5a1fca1df29ebe028df02473966ed723d06e5a/PCB-Potentiometer-Mounts-24-LED-Channels/interactive_bom.json Command failed: /app/src/tasks/processIBOM/run_ibom '/data/checkout/tester/led-zappelin/PCB/Potentiometer_Mounts/16_LED_Channels/Potentiometer_mount_16LED.kicad_pcb' 'PCB-Potentiometer-Mounts-24-LED-Channels' '24 LED channels option board for the LED Zappelin' project' '/data/files/tester/led-zappelin/dc5a1fca1df29ebe028df02473966ed723d06e5a/PCB-Potentiometer-Mounts-24-LED-Channels/interactive_bom.json'
kitspace-processor-1  | /bin/sh: 1: Syntax error: Unterminated quoted string
```

AbdulrhmnGhanem:
> 3. IBOM processing fails with the following error
```console
kitspace-processor-1  | failed tester/led-zappelin/dc5a1fca1df29ebe028df02473966ed723d06e5a/PCB-Potentiometer-Mounts-24-LED-Channels/interactive_bom.json Command failed: /app/src/tasks/processIBOM/run_ibom '/data/checkout/tester/led-zappelin/PCB/Potentiometer_Mounts/16_LED_Channels/Potentiometer_mount_16LED.kicad_pcb' 'PCB-Potentiometer-Mounts-24-LED-Channels' '24 LED channels option board for the LED Zappelin' project' '/data/files/tester/led-zappelin/dc5a1fca1df29ebe028df02473966ed723d06e5a/PCB-Potentiometer-Mounts-24-LED-Channels/interactive_bom.json'
kitspace-processor-1  | /bin/sh: 1: Syntax error: Unterminated quoted string
`````
Fixed in #407.

PR #407, opened 2022-08-30
Escape repo description when running `run_ibom` script

### LINKED ISSUE #406: Prevent shell injection when processing IBOM
Using the right summary as a user can execute a shell injection attack

https://github.com/kitspace/kitspace-v2/blob/4cf2fc47596e4d13765ed13be3d061a84664e8e5/processor/src/tasks/processIBOM/index.ts#L61-L65

AbdulrhmnGhanem:
I tried making `util.exec` escapes input by default, though the `inkscape` command doesn't handle quoted file paths properly.

##
Solves the third issue in #376, and fixes #406 too.


e2e/cypress/integration/syncRepo.spec.js
```diff
@@ -119,12 +119,43 @@ describe('Syncing a project behavior validation', () => {
     cy.get('[data-cy=info-bar]', { timeout: 60_000 }).should('be.visible')
 
     // Go to user projects page to limit the results to the project synced above
```
kasbah, line 121: Comment is wrong now.

PR #410, opened 2022-09-01
Fix race condition in the `sync` function

AbdulrhmnGhanem:
- Debug: I think the flakiness is caused by a race condition in the `sync` fuction
- Deps: add `async-lock`
- Fix: prevent race condition in `sync(gitDir, checkoutDir)`

### DIFF
--- a/processor/src/queues.ts
+++ b/processor/src/queues.ts
@@ -9,8 +9,7 @@ import { JobData } from './jobData'
 import { exists, exec, readFile } from './utils'
 import { DATA_DIR } from './env'
 import redisConnection from './redisConnection'
-import { execSync } from 'child_process'
-import { existsSync } from 'fs'
+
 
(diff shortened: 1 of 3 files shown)


processor/src/queues.ts
```diff
@@ -171,21 +173,33 @@ function tryReadFile(filePath) {
...
+    if (await exists(checkoutDir)) {
+      log.debug('Pulling updates for', gitDir)
+      await exec(`cd ${checkoutDir} && git pull`).catch(err => {
+        // repos with no branches yet will create this error
+        if (
+          err.stderr ===
+          "Your configuration specifies to merge with the ref 'refs/heads/master'\nfrom the remote, but no such ref was fetched.\n"
+        ) {
+          log.warn('repo without any branches', checkoutDir)
+          done(null, err)
```
kasbah, line 190:
```suggestion
          done()
```

PR #411, opened 2022-09-01
Show processor logs in CI if the `test_processor` workflow fails

AbdulrhmnGhanem:
- CI: Show processor logs in CI if the `test_processor` workflow fails
- debug: make the processor tests fail deliberately


AbdulrhmnGhanem:
Fixed with
 https://github.com/kitspace/kitspace-v2/blob/90ef167a0ee5be041fdcf79b9e896f3950bf90e3/processor/test/integration/projects.test.ts#L37-L47
in fbc06824916593de814661c3fc41976c510cbd8c.

PR #417, opened 2022-09-07
Migrate to tsc

AbdulrhmnGhanem:


AbdulrhmnGhanem:
I used `chalk` which is an ES module in `server.ts` and it works without any problem.
https://github.com/kitspace/kitspace-v2/blob/f545f23efcf6d1df4154425c9b26ae0cc00cc610/processor/src/server.ts#L14

processor/src/app.ts
```diff
@@ -1,8 +1,8 @@
 import express from 'express'
 
-import { createProjectsAPI } from './projectAPI'
-import { createRemoteAPI } from './remoteAPI'
-import { createWorkers } from './workers'
+import { createProjectsAPI } from './projectAPI.js'
+import { createRemoteAPI } from './remoteAPI.js'
+import { createWorkers } from './workers.js'
```
kasbah, line 5: I ran into this as well and didn't like this solution. It seems so strange to add .js at the end and it breaks a bunch of stuff since that file doesn't exist a lot of the time. Is there really no better way?

processor/package.json
```diff
@@ -2,23 +2,27 @@
...
   "scripts": {
-    "dev": "nodemon src/server.ts --watch src/",
-    "start": "ts-node src/server.ts",
-    "test": "LOG_LEVEL=debug DATA_DIR=/data/test mocha test/integration/**/*.ts -r ts-node/register --timeout=120000 --exit",
-    "posttest": "rm -rf /data/test",
-    "tsc": "tsc",
+    "dev": "tsc-watch --noClear --onSuccess 'bash -c \"yarn cp-assets && node ./dist/src/server.js\"'",
+    "cp-assets": "find ./src ! -name '*.ts' -exec cp -r '{}' ./dist/src/ \\;",
+    "cp-test-assets": "find ./test ! -name '*.ts' -exec cp -r '{}' ./dist/test/ \\;",
+    "test": "mocha dist/test/integration/**/*.js --timeout=120000 --exit",
```
kasbah, line 16: I really wanted to stick to running the integration tests on the .ts files. Now we will have to mentally map the .js source file location to the .ts file location when we get a stack trace.

kasbah (commented):
I have some quibbles with this but trying to actually address them may not be worth the trouble. Let me know your thoughts anyway. I wish running TS in Node was more straightforward.

processor/src/app.ts
AbdulrhmnGhanem: I don't like too but there are no [other options](https://devblogs.microsoft.com/typescript/announcing-typescript-4-7-beta/#ecmascript-module-support-in-node-js). 

image

processor/package.json
kasbah: A thought I had when I was working on this is to seperate out the integration tests into it's own container. Then run them either on the dev or production server as needed.

AbdulrhmnGhanem: In 9552d89, I added support for [source maps](https://www.typescriptlang.org/tsconfig#sourceMap).

|before|after|
|---|---|
|image|image|

processor/src/server.ts
```diff
@@ -1,15 +1,17 @@
...
+import { giteaDB } from './giteatDB.js'
 
 log.setDefaultLevel((process.env.LOG_LEVEL as log.LogLevelDesc) || log.levels.INFO)
 
 const app = createApp('/gitea-data/git/repositories', { giteaDB })
 
 const port = 5000
 app.listen(port, () => {
-  log.info(`processor listening on http://localhost:${port}`)
+  log.info(chalk.red(`processor listening on http://localhost:${port}`))
```
kasbah, line 14: Can you remove the chalk dependency again please. Doesn't seem like we are actually using it for much.

docker-compose.yml
```diff
@@ -76,11 +76,13 @@ services:
     depends_on:
       - postgres
       - redis
+      - meilisearch
     volumes:
       - gitea-data:/gitea-data
       - processor-data:/data
       # share docker daemon when running docker inside docker
       - /var/run/docker.sock:/var/run/docker.sock
+    tty: true # for colorful logs
```
kasbah, line 85: No more colorful logs ;)

kasbah:
```suggestion
```

PR #443, opened 2022-10-03
Update gitea@193f408

AbdulrhmnGhanem:


AbdulrhmnGhanem:
Getting this error when uploading a file from gitea UI
image

kasbah:
Looks like you found the variable it needs but it looks like a secret to me and hence should go in the `.env`. 

I think we should stop doing major updates to Gitea at the next release (we should still pull in patch updates). We have spent too much time on the updates and need to focus on deploying and using our alpha. We do need to wait for the next Gitea release to stop so this update is ok.

docker-compose.yml
```diff
@@ -19,12 +19,13 @@ services:
       - USER_UID=1000
       - USER_GID=1000
       - SECRET_KEY=${GITEA_SECRET_KEY}
```
kasbah, line 21: It might be replacing this with `...__INTERNAL_TOKEN`?

PR #447, opened 2022-10-11
Replace slugify sub project names with `formatAsGiteaRepoName`

### LINKED ISSUE #442: Make slugify for multi project names the same as Gitea allowed names
E.g. underscores should not be converted (maybe other characters too?)

AbdulrhmnGhanem:
Fixes #442

### DIFF
--- a/processor/src/queues.ts
+++ b/processor/src/queues.ts
@@ -3,8 +3,6 @@ import bullmq from 'bullmq'
 import jsYaml from 'js-yaml'
 import log from 'loglevel'
 import path from 'node:path'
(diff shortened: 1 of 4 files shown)


processor/src/queues.ts
```diff
@@ -3,8 +3,6 @@ import bullmq from 'bullmq'
 import jsYaml from 'js-yaml'
 import log from 'loglevel'
 import path from 'node:path'
-import slugify from 'url-slug'
```
kasbah, line 6: Should it be removed from package.json?

Issue #450, opened 2022-10-13
Review processor is restarting because of git credentials


kasbah:
```
fatal: unable to auto-detect email address (got 'root@cfd641503233.(none)')

    at ChildProcess.exithandler (node:child_process:400:12)
    at ChildProcess.emit (node:events:513:28)
    at maybeClose (node:internal/child_process:1093:16)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:302:5)
```

Can you take a look please @AbdulrhmnGhanem. You can ssh in via deploy@review.staging.kitspace.dev

kasbah:
Looks like it was resolved but did you figure out why it happened?

AbdulrhmnGhanem:
I believe it's a side effect of something else; when I run `git config --list` in a working container there's no username or email both globally and per repo.
Though I couldn't get a shell into the container as it keeps crashing.

kasbah:
Old issue, closing.

PR #451, opened 2022-10-13
Stop running `frontend` and `processor` with root user

AbdulrhmnGhanem:
While investigating #450 I noticed the server runs as `root` which can lead to [some issues](https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/security/non-root-user.md).

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -40,6 +40,7 @@ services:
       dockerfile: Dockerfile
 
   processor:
+    user: root
     build:
       context: processor/
       dockerfile: Dockerfile
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -22,12 +22,14 @@ COPY yarn.lock .
 RUN yarn --frozen-lockfile
 
 FROM base AS production
-COPY --from=build /build/.next/ .next/
-COPY --from=build /deps/node_modules/ node_modules/
-COPY package.json .
-COPY next.config.js .
-COPY public/ public/
-RUN mkdir src/
-COPY src/server.js src/server.js
-ENV NODE_ENV=production
-CMD yarn start
+COPY --chown=node:node --from=build /build/.next/ .next/
+COPY --chown=node:node --from=build /deps/node_modules/ node_modules/
+COPY --chown=node:node package.json .
+COPY --chown=node:node next.config.js .
+COPY --chown=node:node public/ public/
+RUN mkdir src/ &&  chown -R node src
+COPY --chown=node:node src/server.js src/server.js
+ENV NODE_ENV=productionu
+USER node
+
+CMD ["node", "src/server.js"]
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -22,14 +22,20 @@ COPY yarn.lock .
 RUN yarn --frozen-lockfile
 
 FROM base AS production
+
+ENV NODE_ENV=production
+# give the node user read-only permissions
+ARG PERMISSION=644
+COPY --chmod=${PERMISSION} --from=build /deps/node_modules/ node_modules/
+COPY --chmod=${PERMISSION} package.json .
+COPY --chmod=${PERMISSION} next.config.js .
+COPY --chmod=${PERMISSION} public/ public/
+COPY --chmod=${PERMISSION} src/server.js server.js
+
+# The `.next` directory is used by `next/image` to cache optimized images.
+# So it needs to be owned by the `node` user.
 COPY --chown=node:node --from=build /build/.next/ .next/
-COPY --chown=node:node --from=build /deps/node_modules/ node_modules/
-COPY --chown=node:node package.json .
-COPY --chown=node:node next.config.js .
-COPY --chown=node:node public/ public/
-RUN mkdir src/ &&  chown -R node src
-COPY --chown=node:node src/server.js src/server.js
-ENV NODE_ENV=productionu
+
 USER node
 
-CMD ["node", "src/server.js"]
+CMD ["node", "server.js"]
--- a/processor/.dockerignore
+++ b/processor/.dockerignore
@@ -2,3 +2,4 @@
 *Dockerfile*
 *docker-compose*
 node_modules
+dist
--- a/processor/Dockerfile
+++ b/processor/Dockerfile
@@ -19,14 +19,18 @@ RUN apt-get update -qq && apt-get install -qq --no-install-recommends -y \
     python3-pip \
     git
 
-RUN mkdir /app
 WORKDIR /app
-
 COPY . .
-
 RUN pip3 install -r requirements.txt
 RUN yarn install
 RUN yarn tsc
 RUN yarn cp-assets
 
+RUN addgroup --gid 1000 node && \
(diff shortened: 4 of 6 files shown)


processor/.dockerignore
```diff
@@ -2,3 +2,4 @@
 *Dockerfile*
 *docker-compose*
 node_modules
+dist
```
kasbah, line 5: What's in dist?

processor/Dockerfile
```diff
@@ -20,14 +20,19 @@ RUN apt-get update -qq && apt-get install -qq --no-install-recommends -y \
...
 RUN yarn install
 RUN yarn tsc
 RUN yarn cp-assets
 
+RUN addgroup --gid 1000 node && \
+    adduser -u 1000 --gid 1000 node --shell /bin/bash --home /home/node && \
+    mkdir /data /gitea-data && \
+    chown -R node /data /gitea-data && \
+    git config --global user.name "processor" && \
+    git config --global user.email "processor@kitspace.org"
```
kasbah, line 35: We shouldn't need this. When we got errors about the lack of these something else must have been going wrong. I'd prefer not add this unless we actually want to use it.

processor/.dockerignore
AbdulrhmnGhanem: Compiled js from ts files

kasbah: Isn't that exactly what *should* go in the docker container?

AbdulrhmnGhanem: .dockerignore is used when we copy files from the host to the container, the `dist` dir gets generated inside the container itself when compiling. Copying the host `dist` dir causes permission problems (and will cause problems if the container is running on windows too).

kasbah: Ok, thanks for explaining.

docker-compose.override.yml
```diff
@@ -40,6 +40,7 @@ services:
       dockerfile: Dockerfile
 
   processor:
+    user: root
```
kasbah, line 43: Do we really need to run as root for development? Seems like you chown everything to the node user that we would need. Is it because the mounted volumes are owned by root?

frontend/Dockerfile
```diff
@@ -22,12 +22,14 @@ COPY yarn.lock .
...
-ENV NODE_ENV=production
-CMD yarn start
+COPY --chown=node:node --from=build /build/.next/ .next/
+COPY --chown=node:node --from=build /deps/node_modules/ node_modules/
+COPY --chown=node:node package.json .
+COPY --chown=node:node next.config.js .
+COPY --chown=node:node public/ public/
+RUN mkdir src/ &&  chown -R node src
+COPY --chown=node:node src/server.js src/server.js
+ENV NODE_ENV=productionu
```
kasbah, line 32:
```suggestion
ENV NODE_ENV=production
```

docker-compose.override.yml
AbdulrhmnGhanem: chown'ing in the frontend is doable because we are using a multi-stage docker file, but in the processor, the same stage is used for production and development.

I made the user root in development mode because the node user doesn't own the `src` dir, so recompiling and reloading the server on file save doesn't work.

kasbah: Ok, lets fix that when we make processor Dockerfile multi-stage.

frontend/Dockerfile
```diff
@@ -22,12 +22,14 @@ COPY yarn.lock .
...
-COPY public/ public/
-RUN mkdir src/
-COPY src/server.js src/server.js
-ENV NODE_ENV=production
-CMD yarn start
+COPY --chown=node:node --from=build /build/.next/ .next/
+COPY --chown=node:node --from=build /deps/node_modules/ node_modules/
+COPY --chown=node:node package.json .
+COPY --chown=node:node next.config.js .
+COPY --chown=node:node public/ public/
```
kasbah, line 29: I'm wondering if the node user needs to chown all this stuff now. It shouldn't need to write to any of it?

AbdulrhmnGhanem: You are probably right, it only needs read permission.

AbdulrhmnGhanem: I replaced ownership with read/execute permission in f49d18d.

frontend/Dockerfile
```diff
@@ -22,12 +22,20 @@ COPY yarn.lock .
...
-COPY package.json .
-COPY next.config.js .
-COPY public/ public/
-RUN mkdir src/
-COPY src/server.js src/server.js
+
 ENV NODE_ENV=production
-CMD yarn start
+# Make files are readable execuatable bu the `node` user.
+ARG PERMISSION=0555
```
kasbah, line 28:
```suggestion
# allow the node user to read files and nothing else
ARG PERMISSION=004
```

frontend/Dockerfile
```diff
@@ -22,12 +22,20 @@ COPY yarn.lock .
...
+COPY --chmod=${PERMISSION} --from=build /deps/node_modules/ node_modules/
```
kasbah, line 29: Does the chmod get applied recursively?

frontend/Dockerfile
AbdulrhmnGhanem: I committed it to see whether it works or not but I guess it won't work because I tried 444 before and it didn't work.

AbdulrhmnGhanem: Oops, it worked :sweat_smile:, appearently it failed because of the `.next` directory.

kasbah: I guess 004 is a bit weird (root owns it but can't read it?) let's go with 444.

frontend/Dockerfile
kasbah: ?

AbdulrhmnGhanem: I couldn't find any mention in the docs but I running `stat` in the containers shows it's recursive
image

kasbah: That says 755?

AbdulrhmnGhanem: yes this seems wrong

AbdulrhmnGhanem: It's different for the package.json file
image

kasbah: These screenshots are really not helping me figure this out. Let me know when you have an answer.

kasbah:
```
COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker-compose build
```
https://www.docker.com/blog/faster-builds-in-compose-thanks-to-buildkit-support/

frontend/Dockerfile
```diff
@@ -22,12 +22,20 @@ COPY yarn.lock .
...
-COPY package.json .
-COPY next.config.js .
-COPY public/ public/
-RUN mkdir src/
-COPY src/server.js src/server.js
+
 ENV NODE_ENV=production
-CMD yarn start
+# allow the node user to read files and nothing else
+ARG PERMISSION=004
```
kasbah, line 28:
```suggestion
# give the node user read-only permissions
ARG PERMISSION=644
```

kasbah:
Could you resolve the merge conflict with #458 so it will deploy to review.staging please?

PR #453, opened 2022-10-17
Update gitea to 1.17.3+d6a0327

kasbah:
Gitea did a release 2 days ago. There's a few commits since then that are included too but let's stick with this version for now.


kasbah:
Gravatar fetching seems to be broken by this but otherwise it seems fine so far. We don't really need gravatars yet.

PR #466, opened 2022-10-27
Add  `ConflictModal` and `SyncConflictModal` components

### LINKED ISSUE #441: Name conflict on import/sync repo has no resolution
image

AbdulrhmnGhanem:
Fixes #441


kasbah:
One last thing maybe worth trying is to install `@testing-library/jest-dom` alongside `@test-library/user-events` and just get some of the examples from user-events to work without changing them. Then go from there...

Seems you can use `jest-dom` without `jest`: https://www.linkedin.com/pulse/snippet-vite-react-vitest-testing-library-lucian-crisan-

frontend/src/components/NewProject/Sync/index.jsx
```diff
@@ -6,22 +6,25 @@ import { useRouter } from 'next/router'
...
 import { AuthContext } from '@contexts/AuthContext'
-import { mirrorRepo } from '@utils/giteaApi'
-import { SyncOp, NoOp } from '../Ops'
+import { deleteRepo, mirrorRepo } from '@utils/giteaApi'
 import { urlToName } from '@utils/index'
-import styles from './index.module.scss'
 import SyncRepoFromModel from '@models/SyncRepoForm'
 import useForm from '@hooks/useForm'
+import { SyncOp, NoOp } from '../Ops'
+import { SyncConflictModal } from '../ConflictModals'
```
kasbah, line 14: Let's not start pluralizing or component modules, even if they do contain more than one component.

frontend/src/utils/index.js
```diff
@@ -37,7 +36,12 @@ export const b64toBlob = base64 => fetch(base64).then(res => res.blob())
  * urlToName('https://github.com/emard/ulx3s/')
  */
 export const urlToName = url => {
-  const urlObj = new URL(url)
+  let urlObj
+  try {
+    urlObj = new URL(url)
+  } catch (e) {
+    return ''
```
kasbah, line 43: What happens in the form when we return the empty string?

frontend/src/components/NewProject/Sync/index.jsx
```diff
@@ -76,46 +76,153 @@ const Sync = ({ setUserOp }) => {
...
+      color: 'green',
+    })
+
+    const repoURL = form.url
+    const repoName = urlToName(repoURL)
+
+    const deletedRepoSuccessfully = await deleteRepo(
+      `${user.username}/${projectName}`,
+      apiToken,
+    )
```
kasbah, line 131: Deleting a repo should always be behind several confirmations. I feel like we shouldn't be offering the overwrite option on migration at all. The choice stems from the upload flow where we imagined a user might regularly upload a new file to update the project. 

Since we won't be letting users migrate repos themselves on alpha-2 I'd be happy to defer changing this to the next milestone though.

kasbah (changes requested):
- Looks good overall except for some repeated logic
- Nice job on the tests and getting them working
- We shouldn't offer overwriting I think, but happy for us to open an issue for that and defer changing it to the next milestone.

frontend/package.json
```diff
@@ -58,9 +59,10 @@
     "eslint-plugin-import": "^2.23.4",
     "eslint-plugin-prettier": "3.4.0",
     "jsdom": "^20.0.1",
+    "next-router-mock": "^0.7.4",
     "prettier": "^2.0.5",
     "sass": "^1.49.6",
-    "typescript": "~4.3.5",
+    "typescript": "^4.8.4",
```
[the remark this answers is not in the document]
kasbah:
```suggestion
    "typescript": "~4.8.4",
```

kasbah:
Oh and can you give this PR a better name please.

e2e/cypress/integration/uploadProject.spec.js
```diff
@@ -28,7 +28,7 @@ describe('Upload project', () => {
     })
 
     cy.get('[data-cy=creating-project-loader]')
-    cy.url().should('eq', `${Cypress.config().baseUrl}/${username}/example`)
+    cy.url().should('eq', `${Cypress.config().baseUrl}/${username}/example.png`)
```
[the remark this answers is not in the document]
AbdulrhmnGhanem: `example.png` is a valid project name now. 

```diff
export const slugifiedNameFromFiles = files => {
   // ...
-  return slugify(projectWithExt.split('.')[0])
+  return formatAsGiteaRepoName(projectWithExt)
}
```

frontend/src/components/NewProject/Sync/index.jsx
AbdulrhmnGhanem: Resolved in 20a9189.

frontend/src/components/NewProject/Sync/index.jsx
```diff
@@ -76,46 +76,153 @@ const Sync = ({ setUserOp }) => {
...
+
+    const res = await mirrorRepo(repoURL, uid, apiToken, repoName)
+    const migrateSuccessfully = res.ok
+
+    if (migrateSuccessfully) {
+      setMessage({
+        content: 'Migrated successfully, redirecting the project page...',
+        color: 'green',
+      })
+      await push(`/${username}/${repoName}`)
```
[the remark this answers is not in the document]
AbdulrhmnGhanem: Resolved in 4ea980b.

frontend/src/components/NewProject/Sync/index.jsx
AbdulrhmnGhanem: #490

kasbah:
> Add HOC ConflictModal and SyncConflictModal
Does HOC stand for higher order component? So a component that takes in another component as an argument and transforms it? Do `ConflictModal` and `SyncConflictModal` do that?

AbdulrhmnGhanem:
We are using `ConflictModal` to create `SyncConflictModal` and `UploadConflictModal`.
I edited the name to clarify that `SyncConflictModal` is a normal component.

frontend/src/components/NewProject/ConflictModal.tsx
```diff
@@ -0,0 +1,166 @@
+import React, { useContext, useEffect } from 'react'
+import { Button, Form, Input, Modal } from 'semantic-ui-react'
+
+import UseForm from '@hooks/useForm'
```
kasbah, line 4: I think this should be lowercase `useForm`?

frontend/src/components/NewProject/ConflictModal.tsx
```diff
@@ -0,0 +1,166 @@
...
+const ConflictModal = ({
+  conflictModalOpen,
+  onClose,
+  originalProjectName,
+  message,
+  onDifferentName,
+  onDifferentNameButtonContent,
+  onOverwrite,
+  onOverwriteButtonContent,
+}: BaseConflictModalProps) => {
```
kasbah, line 19: I'm not too familiar with the higher-order-component terminology but it is doesn't look like a HOC to me. A HOC from [React docs](https://reactjs.org/docs/higher-order-components.html) is something like:

```js
const EnhancedComponent = higherOrderComponent(WrappedComponent);
```

frontend/src/components/NewProject/ConflictModal.tsx
AbdulrhmnGhanem: yes

frontend/src/components/NewProject/ConflictModal.tsx
AbdulrhmnGhanem: I don't know what to call it :laughing: it\s what react calls [specialization](https://reactjs.org/docs/composition-vs-inheritance.html#specialization), maybe just call both of them components.

frontend/src/utils/index.ts
```diff
@@ -30,15 +29,20 @@ export const b64toBlob = base64 => fetch(base64).then(res => res.blob())
...
  */
 export const urlToName = url => {
-  const urlObj = new URL(url)
-  return path.basename(urlObj.pathname, path.extname(urlObj.pathname))
+  try {
+    url = new URL(url)
+  } catch {
+    return ''
+  }
+  return path.basename(url.pathname, path.extname(url.pathname))
```
kasbah, line 45: I thought you changed this while we were pair programming?

AbdulrhmnGhanem: The cleanest way to check if a string is a valid URL is to construct a URL object.

Checking if the value is falsey or empty string won't work, because a non-empty string would pass that check and cause an exception when calling `new URL(val)`.
We didn't face this problem because we were copy-pasting the URL, so we never called this function with a string which is invalid URL.

Issue #269, opened 2022-02-02
Mix of styled-jsx and scss modules


kasbah:
The original justification for adding `styled-jsx` (9a4c6c6f94cf637e09ad6bd92d673c1076192bf4) was that it's needed for `dangerouslySetInnerHTML` which we only use for the readme. Now it's being used all over the place and being mixed it with scss modules. We should pick one or the other and use that, i.e. one of:
- Use it just for the readme and scss modules everywhere else
- Use `styled-jsx` everywhere
  
Any preference @AbdulrhmnGhanem ?

(Also using `<style global` everywhere. What's the point of breaking out of the component scope all the time?)

AbdulrhmnGhanem:
`styled-jsx` is used for styling elements that aren't under our control for example the readme and the parts of semantic-ui which aren't customizable via `scss`.
`global` has been used for the same reason, these elements aren't under the component control for example

|w/ global|w/o global|
|---|---|
|![image](https://user-images.githubusercontent.com/37152329/152182847-9d69a3a1-969e-4880-bbcf-0aa2391e6892.png)|![image](https://user-images.githubusercontent.com/37152329/152182558-7b9e3858-06ee-4747-9785-e8803dd2301d.png)|

We can move the jsx styles into `_app.scss` but I prefer keeping the style in the parent component.

kasbah:
Ok, when I said "all over the place" I forgot to actually check. Seems like that isn't true. :rofl: It's used twice in `NavBar` though. 

The readme makes sense I think (maybe there's a better way?). The one I was looking at is this: it sets an id in the component and then a `<style global jsx` to zero back in on the id modify it? That doesn't make a whole lot of sense. Is that really necessary?
https://github.com/kitspace/kitspace-v2/blob/ae711eb45e986fbfe19d3d57e98c4d950913eae8/frontend/src/components/NavBar/index.js#L91-L106

AbdulrhmnGhanem:
Yes this one doesn't need `global`.

kasbah:
And could be set from `module.scss`?

AbdulrhmnGhanem:
> And could be set from module.scss?

Yes, c9a15a76df43d250bfb1af69f66e7705ebf0a0d7.

kasbah:
Since it confused me once again, let's stop using it in the menu and add a comment to the readme one.

PR #487, opened 2022-11-05
Use swc in frontend

kasbah:
I think it's time. Rebases #265 and gives the branch a better name.


AbdulrhmnGhanem:
wdyr got supported by swc now https://github.com/vercel/next.js/issues/38502

kasbah:
Ah, good to know. Let's add it when we need it on another PR?

AbdulrhmnGhanem:
And jsx styles work out of the box now, great!

AbdulrhmnGhanem (approved):

kasbah:
We don't seem to actually using styled-jsx. Maybe I should have just removed it?

kasbah:
Ah we are using it in Readme.jsx, but it's `global` anyway? Could we just use .module.scss like the rest?

AbdulrhmnGhanem:
We are

image
image

https://github.com/kitspace/kitspace-v2/blob/2f27987f200e0441f5ee6a0972167d65049be83d/frontend/src/components/Board/Readme.jsx#L10-L66

and for the mobile nav bar too.

kasbah:
Can you remember why we didn't use module.scss for these though?

AbdulrhmnGhanem:
#269

PR #489, opened 2022-11-06
Update nextjs to latest v12.*

kasbah:
- Updates nextjs to latest v12 (v13 is out but there are breaking changes)
- Had to update linting
   - While I was at it I made `_` `__` `___` etc. the allowed ignored unused vars


kasbah:
Getting weird (transpilaton?) errors with the upgrade and the return of `getKitspaceYamlArray`, it doesn't return an empty array on 404.

PR #492, opened 2022-11-07
meilisearch: make document id compliant with the spec

AbdulrhmnGhanem:
Projects with a `.` in the name  (e.g., `sangaboard_v0.3`) weren't getting added to meilisearch.

### DIFF
--- a/processor/src/tasks/addToSearch.ts
+++ b/processor/src/tasks/addToSearch.ts
@@ -50,7 +50,10 @@ export default async function addToSearch(
     return
   }
 
-  const searchId = `${giteaId}-${subprojectName}`
+  // A document identifier must be of type integer or string,
+  // composed only of alphanumeric characters (a-z A-Z 0-9), hyphens (-), and underscores (_).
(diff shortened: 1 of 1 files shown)


processor/src/tasks/addToSearch.ts
```diff
@@ -51,6 +51,10 @@ export default async function addToSearch(
   }
 
   const searchId = `${giteaId}-${subprojectName}`
+    // A document identifier must be of type integer or string,
+    // composed only of alphanumeric characters (a-z A-Z 0-9), hyphens (-), and underscores (_).
+    // See, https://docs.meilisearch.com/reference/errors/error_codes.html#invalid-document-id
+    .replace(/[^\w\d-_]/g, '-')
```
kasbah, line 57:
```suggestion
  // A document identifier must be of type integer or string,
  // composed only of alphanumeric characters (a-z A-Z 0-9), hyphens (-), and underscores (_).
  // See, https://docs.meilisearch.com/reference/errors/error_codes.html#invalid-document-id
  const searchId = `${giteaId}-${subprojectName.replace(/[^\w\d-_]/g, '-')}`
```

PR #495, opened 2022-11-11
Fix e2e names and emails and make validation rules for `Sign{Up,In}Form` match gitea

### LINKED ISSUE #481: e2e generates possibly real email addresses
If we ever accidentally hook up an email server while testing it will try and send a bunch of emails. Most likely outcome is we will get a lot of bounces and get blocked because of it. We might even email some real people by mistake. 

We should use `@example.com` addresses which is made for this purpose, has no MX record and the SMTP port isn't open so will prevent further harm. (https://serverfault.com/a/333552)

### LINKED ISSUE #482: e2e can generate invalid username
image

AbdulrhmnGhanem:
- e2e: use faker `userName` instead of `firstName` and @example emails.
- Make validation rules for `Sign{Up,In}Form` match gitea.
##
Fixes #482 and closes #481

### DIFF
--- a/e2e/cypress/integration/IBOM.spec.js
+++ b/e2e/cypress/integration/IBOM.spec.js
@@ -1,6 +1,4 @@
-import faker from 'faker'
-
-import { getFakeUsername } from '../support/getFakeUsername'
+import { getFakeUser } from '../support/getFakeUser'
 
 describe('IBOM page', () => {
   before(() => {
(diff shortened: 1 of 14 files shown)


e2e/cypress/support/getFakeUsername.js
```diff
@@ -2,18 +2,22 @@ import faker from 'faker'
 
 /**
  * `faker.unique` isn't that unique; sometimes it returns a previously used username.
- * @returns {string} unique username
+ * @returns {{username: string, email: string, password: string}}}}}
  */
-export const getFakeUsername = () => {
+export const getFakeUser = () => {
```
kasbah, line 7: I think the file should be renamed too.

PR #496, opened 2022-11-11
Remove styled-jsx

### LINKED ISSUE #269: Mix of styled-jsx and scss modules

AbdulrhmnGhanem:
- Refactor: move SmallMenu style into its SCSS module
- Refactor: replace styled-jsx with scss file for items alignment
- Refactor: replace styled-jsx with scss file for readme
- Deps: remove styled-jsx
##
|before (menu)|after (menu)|
|---|---|
|image|image|

|before (readme)|after (readme)|
|---|---|
|image|image|

##
Closes #269.

### DIFF
--- a/frontend/src/components/Board/Readme.jsx
+++ b/frontend/src/components/Board/Readme.jsx
@@ -7,7 +7,6 @@ const Readme = ({ renderedReadme }) => (
       // eslint-disable-next-line react/no-danger
       dangerouslySetInnerHTML={{ __html: renderedReadme }}
     />
-    {/* <style global jsx>{``}</style> */}
   </div>
 )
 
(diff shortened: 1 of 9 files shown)


frontend/src/components/Board/Readme.jsx
```diff
@@ -7,62 +7,7 @@ const Readme = ({ renderedReadme }) => (
...
-      #readme table td {
-        padding: 6px 13px;
-        border: 1px solid #d0d7de;
-      }
-
-      #readme table tr:nth-child(2n) {
-        background-color: #f6f8fa;
-      }
-    `}</style>
+    {/* <style global jsx>{``}</style> */}
```
kasbah, line 10:
```suggestion
```

kasbah:
I don't like that button centering in the menu.

PR #504, opened 2022-11-23
Lazy infinite loading of project cards

AbdulrhmnGhanem:
- UI: add styles for the `CardsGrid` component
- Load projects lazily in the `CardsGrid` component
- Refactor: move `ProjectsGrid` into a stand-alone component
- Use `ProjectsGrid` component for user projects
- Use `ProjectsGrid` component for multi projects page
##

### Lazy loading

https://user-images.githubusercontent.com/37152329/203544524-7faf4b83-4686-4612-b099-148238dff809.mp4

### Homepage
image
##

### User projects
image
##

### Multi project page
image


frontend/src/components/CardsGrid/index.module.scss
```diff
@@ -0,0 +1,22 @@
+@import '../../components/ProjectCard/index.module.scss';
+
+$gap: 16px;
+
+.cardsGrid {
+  /* 
+    We can't simply use a css grid because it doesn't support centering elements if there are less
+    than 3 elements in the last row.
+  */
```
kasbah, line 9:
```suggestion
```

I don't think this comment is necessary. Is there really a good reason to prefer `grid` over `flex` here? Grid always seemed more complicated to me, but I've haven't learned it yet. 

Though regarding the centering, I'm also not sure it looks good when there are only a few projects (e.g. on a user page), but we can leave that for now since the user page has not really been styled yet.

frontend/src/components/ProjectCard/index.jsx
```diff
@@ -7,46 +7,52 @@ import Image from 'next/image'
...
+                <Image
+                  alt={`Render of PCB of ${nameOnCard}`}
+                  data-cy="project-card-thumbnail"
+                  height={180}
+                  objectFit="contain"
+                  src={src}
+                  width={240}
+                />
+              )}
+            </div>
```
kasbah, line 38: Seems to me there is an unneeded `<div>` here, or is it used for something?

frontend/src/pages/[user]/index.jsx
```diff
@@ -24,12 +21,16 @@ export const getServerSideProps = async ({ params }) => {
...
+    filter: `ownerName = ${username}`,
+  }
+
+  const hits = await gridFetcher(q)
 
   return {
     props: {
       swrFallback: {
-        [params.username]: hits,
+        [q]: hits,
```
Comment (kasbah) [inline frontend/src/pages/[user]/index.jsx:33]: This key is `'[object Object]'` for all usernames.

frontend/src/components/ProjectCard/index.jsx
```diff
@@ -7,46 +7,52 @@ import Image from 'next/image'
...
-              <Image
-                alt={`Render of PCB of ${nameOnCard}`}
-                data-cy="project-card-thumbnail"
-                height={180}
-                objectFit="contain"
-                src={src}
-                width={240}
-              />
-            )}
+const ProjectCard = forwardRef(
```
kasbah, line 10: Are you sure `forwardRef` is needed here? I think without it the `ref` just gets attached to `<Link>` automatically and this component doesn't need to worry about it.

kasbah (changes requested):
- [x] Looks like a pretty nice solution overall
- [ ] I haven't quite figured out why, but it doesn't seem to be server rendering the card grid, which makes the initial loading experience worse
- [ ] Some minor quibbles:
    - [ ] Your PR title is talking about a component that doesn't exist
    - [x] I think the component should be called `ProjectCardGrid`
    - [ ] You shouldn't really name your PR title for "adding a component" but talk about the feature you are implementing. Why are you adding this component? Also you are doing much more than just adding a component...
    - [x] Could you please start prefixing your branches with your initials like I have been doing with `ke/`. It will make cleaning up branches easier in the future.

frontend/src/pages/[user]/index.jsx
```diff
@@ -50,15 +51,22 @@ UserPage.propTypes = {
...
 
 const User = ({ username }) => {
-  const { data: userProjects } = useSWR(username, fetchUserProjects)
+  const { data, setSize } = useSWRInfinite(
+    getKey('*', `ownerName = ${username}`),
+    gridFetcher,
+  )
+  const intersectionObserverRef = useUpdateBeforeReachingLimit(setSize)
+
+  const userProjects = data?.flat()
```
Comment (kasbah) [inline frontend/src/pages/[user]/index.jsx:60]: Does it need `flat()`? I think it would be good to use typescript types to clear up what's happening here.

frontend/src/pages/search/index.jsx
```diff
@@ -1,25 +1,25 @@
...
-  const searchResult = await meiliIndex.search(query)
-  return searchResult.hits
-}
-
 export const getServerSideProps = async ({ query }) => {
   // '*' or '' means return everything but '' can't be used as a SWR cache key
   const { q = '*' } = query
 
-  const hits = await fetchSearch(q)
+  const hits = await gridFetcher({ query: q })
```
kasbah, line 22: This makes for some confusing reading. 

image

frontend/src/components/ProjectCard/index.jsx
AbdulrhmnGhanem: It's needed if you pass the ref as `<Component ref={value} />` I got the following error without `forwardRef`
image

I got rid of `forwardRef` by passing the ref under a different name :point_down: 
```jsx
<ProjectCard
    intersectionObserverRef={
      index === projects.length - cardsPerRow * 2
        ? intersectionObserverRef
        : null
    }
 />
```

frontend/src/components/ProjectCardGrid/index.module.scss
```diff
@@ -0,0 +1,18 @@
...
+  row-gap: $gap;
+  column-gap: $gap;
+  margin: auto;
+  justify-content: center;
+  /* There are three cards per row, so there are to gaps in each row. */
+  max-width: calc($project-card-width * 3 + $gap * 2);
+}
+
+.cardsGrid > div {
+  flex: 0 0 0 33.333%;
```
kasbah, line 17: I'll never remember what this means. Could you break this down into individual properties?

frontend/src/components/ProjectCardGrid/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const searchResult = await meiliIndex.search(query, { limit, offset, filter })
+  return searchResult.hits as Project[]
+}
+
+/**
+ * Update useSWRInfinite size when the user reaches two rows from the end of the grid.
+ * @param  setSize - the function returned by useSWRInfinite.
+ */
+export const useUpdateBeforeReachingLimit = (
+  setSize: (size: (_size: number) => number) => Promise<Project[][] | undefined>,
```
kasbah, line 79: Please separate the types out to make this more readable.

kasbah: Here's an example: https://github.com/kitspace/kitspace-v2/blob/8fdada80cfb49920ebcef379a587817fc1b37f44/frontend/src/utils/index.ts#L98-L115

frontend/src/pages/[user]/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+  getKey,
+  gridFetcher,
+  Project,
+  useUpdateBeforeReachingLimit,
+} from '@components/ProjectCardGrid'
+import { userExists } from '@utils/giteaApi'
+import { GetServerSideProps } from 'next'
+
+export const getServerSideProps: GetServerSideProps = async ({ params }) => {
+  const username = params.user as string
```
Comment (kasbah) [inline frontend/src/pages/[user]/index.tsx:16]: Why is the `as` needed here? What if it's null?

frontend/src/components/ProjectCardGrid/index.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { useEffect } from 'react'
+import { useInView } from 'react-intersection-observer'
+
+import { meiliIndex } from '@utils/meili'
+import { Filter } from 'meilisearch'
+import ProjectCard, { IntersectionObserverRef } from '../ProjectCard'
+import styles from './index.module.scss'
+
+export const cardsPerRow = 3
+const LIMIT = cardsPerRow * 6
```
kasbah, line 10: Why is LIMIT capitalized?

frontend/src/components/ProjectCardGrid/index.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+export const getKey =
+  ({ query, filter }: { query: string; filter?: Filter }) =>
+  (pageIndex: number, previousPageData: Project[]) => {
+    if (previousPageData && !previousPageData.length) {
+      // reached the end
+      return null
+    }
+    return { query, offset: pageIndex * LIMIT, limit: LIMIT, filter }
+  }
```
kasbah, line 55: The name `getKey` is confusing to me. Key of what? It seems to be returning an object.

frontend/src/pages/[user]/[repo]/index.tsx
```diff
@@ -1,15 +1,18 @@
...
+import React from 'react'
 import { NextPageContext } from 'next'
 import getConfig from 'next/config'
-import React from 'react'
 import useSWR, { SWRConfig, unstable_serialize } from 'swr'
+
+import { getKitspaceYamlArray } from '@utils/projectPage'
+import { getRepo, repoExists } from '@utils/giteaApi'
+import { waitFor } from '@utils/index'
+import CardsGrid, {
```
Comment (kasbah) [inline frontend/src/pages/[user]/[repo]/index.tsx:9]: The import needs to be renamed.

frontend/src/pages/[user]/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+const UserPage = ({ swrFallback, username }: UserPageProps) => {
+  return (
+    <SWRConfig value={{ fallback: swrFallback }}>
+      <User username={username} />
+    </SWRConfig>
+  )
+}
+
+const User = ({ username }: { username: string }) => {
```
Comment (kasbah) [inline frontend/src/pages/[user]/index.tsx:54]: Please separate out the props as a type, even if it's just one.  Maybe there's an eslint rule for this?

frontend/src/pages/[user]/index.tsx
```diff
@@ -0,0 +1,74 @@
...
+    query: '*',
+    filter: `ownerName = ${username}`,
+  }
+
+  const hits = await gridFetcher(q)
+
+  return {
+    props: {
+      swrFallback: {
+        [unstable_serialize(getKey(q))]: hits,
```
Comment (kasbah) [inline frontend/src/pages/[user]/index.tsx:34]: This is still a little confusing. It's called `q` because in the URL it's `?q=` but you are calling that string `query` and it's part of a `q` object.

frontend/src/components/ProjectCardGrid/index.tsx
AbdulrhmnGhanem: SWR keys can be objects, this is why we are using `unstable_serialize`.  And `getKey` is the name of this function in [SWRInfinite itself](https://swr.vercel.app/docs/pagination#parameters).

frontend/src/pages/[user]/index.tsx
Comment (AbdulrhmnGhanem) [reply frontend/src/pages/[user]/index.tsx]: I changed it to `searchParams`.

AbdulrhmnGhanem:
> I haven't quite figured out why, but it doesn't seem to be server rendering the card grid, which makes the initial loading experience worse

Fixing the serialization of the keys made SSR work, here is the HTML returned by the server :point_down: 

image

frontend/src/components/ProjectCardGrid/index.tsx
AbdulrhmnGhanem: Just to get around this https://github.com/kitspace/kitspace-v2/blob/9e229ddea33313dfc0b8d4d7d7bf10b713245cc3/frontend/src/components/ProjectCardGrid/index.tsx

I rename it to `defaultLimit`

frontend/src/components/ProjectCardGrid/index.module.scss
AbdulrhmnGhanem: https://github.com/kitspace/kitspace-v2/pull/504/files#diff-b96cdfa95b55b447ea0c6c06b30aaa71e137d1e4af5c2e39eec5030ade7ba933R17

frontend/src/components/ProjectCardGrid/index.tsx
kasbah: Could you call it something like `getSwrKey` to match `swrFallback`?

frontend/src/components/ProjectCardGrid/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [ref, isReachingLimit] = useInView({ triggerOnce: true })
+
+  useEffect(() => {
+    if (isReachingLimit) {
+      setSize(size => size + 1)
+    }
+  }, [isReachingLimit, setSize])
+
+  return ref
+}
```
kasbah, line 93: I'm wondering about this. The name `useUpdateBeforeReachingLimit` is pretty intimidating and I don't like that `setSize` is passed in like that. Would this be clearer if `useSWRInfinite` and `useInView` were wrapped up in one hook?

```ts
export interface SearchParams {
  query: string
  filter: string
}
export const useSearch = (params: SearchParams) => {
  const [ref, isReachingLimit] = useInView({ triggerOnce: true })
  const { data, setSize } = useSWRInfinite(getKey(params), gridFetcher)

  useEffect(() => {
    if (isReachingLimit) {
      setSize(size => size + 1)
    }
  }, [isReachingLimit, setSize])

  return { projects: data?.flat(), ref }
}
```

AbdulrhmnGhanem: Resolved in 11b9354.

frontend/src/components/ProjectCardGrid/index.tsx
AbdulrhmnGhanem: Resolved in 11b9354.

frontend/src/components/ProjectCardGrid/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+import useSWRInfinite from 'swr/infinite'
+import { useInView } from 'react-intersection-observer'
+import { Filter } from 'meilisearch'
+
+import { meiliIndex } from '@utils/meili'
+import ProjectCard, { IntersectionObserverRef } from '../ProjectCard'
+import styles from './index.module.scss'
+
+export const cardsPerRow = 3
+const defaultLimit = cardsPerRow * 6
```
kasbah, line 11: 1. `cardsPerRow` is exported but not used anywhere?
2. How does this behave when there are less than 3 cards per row, e.g. on mobile?

frontend/src/components/ProjectCardGrid/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+  intersectionObserverRef,
+}: ProjectCardGridProps) => {
+  return (
+    <div className={styles.cardsGrid} data-cy="cards-grid">
+      {projects?.map((project, index) => (
+        <ProjectCard
+          {...project}
+          key={project.id}
+          intersectionObserverRef={
+            index === projects.length - cardsPerRow * 2
```
kasbah, line 37: This could be put into a variable up top and given a nicer name.

kasbah (changes requested):

frontend/src/components/ProjectCardGrid/index.tsx
kasbah: I just noticed this is returning a function so I think it should be called something that makes that clear. `makeSWRInfiniteKeyGetter`?

frontend/src/pages/[user]/index.tsx
```diff
@@ -0,0 +1,70 @@
...
+    query: '',
+    filter: `ownerName = ${username}`,
+  }
+
+  const hits = await searchFetcher(searchParams)
+
+  return {
+    props: {
+      swrFallback: {
+        [unstable_serialize(getSWRKey(searchParams))]: [hits],
```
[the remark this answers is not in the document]
Comment (AbdulrhmnGhanem) [reply frontend/src/pages/[user]/index.tsx:34]: Yes, SWR key can be a function, `useSWRInfinite` takes exactly the same value. And `unstable_serialize` handles it as expected :point_down: 

image

frontend/src/components/ProjectCardGrid/index.tsx
```diff
@@ -0,0 +1,33 @@
+import Project from '@models/Project'
+import ProjectCard, { IntersectionObserverRef } from '../ProjectCard'
+import styles from './index.module.scss'
+
+export const cardsPerRow = 3
+
+interface ProjectCardGridProps {
+  projects: Project[]
```
kasbah, line 8:
```suggestion
  projects: Array<Project>
```

It's somewhat arbitrary but I've been trying to stick to one style. I'll add an eslint rule at some point.

frontend/src/hooks/useLazySearch.ts
```diff
@@ -0,0 +1,58 @@
...
+  limit: number
+  offset: number
+}
+
+export const cardsPerRow = 3
+const defaultLimit = cardsPerRow * 6
+
+export const makeSWRKeyGetter =
+  ({ query, filter }: SearchParams) =>
+  (pageIndex: number, previousPageData: Project[]): SearchSWRKeyParams => {
```
kasbah, line 24:
```suggestion
  (pageIndex: number, previousPageData: Array<Project>): SearchSWRKeyParams => {
```

frontend/src/hooks/useLazySearch.ts
```diff
@@ -0,0 +1,58 @@
...
+export type SearchFetcherParams = Partial<SearchSWRKeyParams>
+
+export const searchFetcher = async ({
+  filter,
+  offset = 0,
+  query = '',
+  limit = defaultLimit,
+}: SearchFetcherParams) => {
+  const searchResult = await meiliIndex.search(query, { limit, offset, filter })
+  return searchResult.hits as Project[]
```
kasbah, line 41:
```suggestion
  return searchResult.hits as Array<Project>
```

frontend/src/pages/[user]/[repo]/index.tsx
```diff
@@ -79,53 +80,47 @@ RepoPage.getInitialProps = async (ctx: NextPageContext): Promise<RepoPageProps>
 }
 
 interface ProjectGridData {
-  swrFallback: Record<string, Array<unknown>>
+  swrFallback: Record<string, Project[]>
```
Comment (kasbah) [inline frontend/src/pages/[user]/[repo]/index.tsx:83]:
```suggestion
  swrFallback: Record<string, Array<Project>>

File: frontend/src/pages/[user]/index.tsx [hunk 23]
```diff
@@ -0,0 +1,71 @@
...
+        [unstable_serialize(makeSWRKeyGetter(searchParams))]: [hits],
+      },
+      username,
+    },
+  }
+}
+
+interface UserPageProps {
+  username: string
+  swrFallback: Record<string, Project[]>
```
Comment (kasbah) [inline frontend/src/pages/[user]/index.tsx:44]:
```suggestion
  swrFallback: Record<string, Project[]>
```

frontend/src/pages/search/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+        [unstable_serialize(makeSWRKeyGetter(searchParams))]: [hits],
+      },
+      initialQuery: query,
+    },
+  }
+}
+
+interface SearchPageProps {
+  initialQuery: string
+  swrFallback: Record<string, Project[][]>
```
kasbah, line 40:
```suggestion
  swrFallback: Record<string, Array<Array<Project>>>
```

kasbah (changes requested):

frontend/src/pages/search/index.tsx
kasbah: Wait, why is this one an array of array but in `[user]/index.tsx` it's just array?

https://github.com/kitspace/kitspace-v2/pull/504/files/53f8def9ded873b0b88fcde04567ecde09175b3e#diff-14d5e273a8f5b377e25c3f2871911a7390c78dd991031efbbf018ef4fe59d294R44

AbdulrhmnGhanem: I didn't type the `props` returned in the `getServerSideProps` return value so it didn't catch it. Fixed it 0085daf.

frontend/src/components/ProjectCardGrid/index.tsx
AbdulrhmnGhanem: I fixed the export. 

We load chunks of 16 projects before we reach the limit.
`cardsPerRow` is used for calculating when we should load more projects. It loads more projects when the viewport reaches multiples of 12: (12, 24, 36, ...) regardless of the number of projects in a row.

frontend/src/pages/search/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const hits = await searchFetcher(searchParams)
+
+  return {
+    props: {
+      swrFallback: {
+        [unstable_serialize(makeSWRKeyGetter(searchParams))]: [hits],
+      },
+      initialQuery: query,
+    } as SearchPageProps,
```
kasbah, line 39: Doesn't this type assertion circumvent type checking?

kasbah (changes requested):

frontend/src/components/ProjectCardGrid/index.tsx
kasbah: Huh? So `cardsPerRow` is used to calculate when to load projects but it loads new projects every 12 regardless of the actual cards per row?

frontend/src/components/ProjectCardGrid/index.module.scss
```diff
@@ -0,0 +1,18 @@
...
+$gap: 16px;
+
+.cardsGrid {
+  display: flex;
+  flex-wrap: wrap;
+  row-gap: $gap;
+  column-gap: $gap;
+  margin: auto;
+  justify-content: center;
+  /* There are three cards per row, so there are to gaps in each row. */
```
kasbah, line 12:
```suggestion
  /* There are three cards per row, so there are two gaps in each row. */
```

frontend/src/pages/[user]/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+    query: '',
+    filter: `ownerName = ${username}`,
+  }
+
+  const hits = await searchFetcher(searchParams)
+
+  return {
+    props: {
+      swrFallback: {
+        [unstable_serialize(makeSWRKeyGetter(searchParams))]: [hits],
```
Comment (kasbah) [inline frontend/src/pages/[user]/index.tsx:40]:
```suggestion
      swrFallback: {
        // unstable_serialize is clever enough to turn our key getter function into the right string key
        [unstable_serialize(makeSWRKeyGetter(searchParams))]: [hits],
```

frontend/src/pages/search/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const query = (queryParams.q as string) || ''
+
+  const searchParams = { query }
+
+  const hits = await searchFetcher(searchParams)
+
+  return {
+    props: {
+      swrFallback: {
+        [unstable_serialize(makeSWRKeyGetter(searchParams))]: [hits],
```
kasbah, line 36:
```suggestion
        // unstable_serialize is clever enough to turn our key getter function into the right string key
        [unstable_serialize(makeSWRKeyGetter(searchParams))]: [hits],
```

frontend/src/pages/[user]/[repo]/index.tsx
```diff
@@ -79,53 +80,47 @@ RepoPage.getInitialProps = async (ctx: NextPageContext): Promise<RepoPageProps>
...
-
 const getGridData = async (
   username: string,
   repoName: string,
 ): Promise<ProjectGridData> => {
   const repoFullName = `${username}/${repoName}`
   const repo = await getRepo(repoFullName)
-  const searchArgs: SearchArgs = ['*', { filter: `repoId = ${repo.id}` }]
-  const hits = await fetchSearch(...searchArgs)
+  const searchParams = { filter: `repoId = ${repo.id}` }
```
Comment (kasbah) [inline frontend/src/pages/[user]/[repo]/index.tsx:116]:
```suggestion
  const searchParams = { filter: `repoId = ${repo.id}`, limit: 1000 }
```

Just to make our limit explicit here. The default is 12 so I fell like it's actually possible to hit that, whereas 1000 doesn't seem likely.

frontend/src/components/ProjectCard/index.tsx
```diff
@@ -20,19 +34,17 @@ const ProjectCard = ({ projectName, summary, ownerName, repoName }) => {
...
-              />
-            )}
-          </div>
+        <div ref={intersectionObserverRef} className={styles.thumbnail}>
+          {isLoading || isError ? null : (
+            <Image
+              alt={`Render of PCB of ${nameOnCard}`}
+              data-cy="project-card-thumbnail"
+              height={180}
+              objectFit="contain"
```
kasbah, line 43:
```suggestion
              objectFit="contain"
              priority={true}
```

At this stage the HTML is loaded, this image is present and in view so it makes sense to load it as quickly as possible.

AbdulrhmnGhanem: I handled this in #515, It's [advisable](https://nextjs.org/docs/api-reference/next/image#priority) to set the priority to `true` for the images that appear in [LCP](https://web.dev/lcp/) in the initial render.

frontend/src/components/ProjectCardGrid/index.tsx
AbdulrhmnGhanem: Yes

frontend/src/pages/search/index.tsx
AbdulrhmnGhanem: Converted it to normal annotation in cedd4fd, the assigned object gets type checked.

frontend/src/components/ProjectCardGrid/index.tsx
kasbah: Ok, can you refactor that to something that's a bit clearer please? I don't think make sense to name this `cardsPerRow`. Maybe `export const pageSize = 12` or something like that, and do away with the calculation?

frontend/src/components/ProjectCard/index.tsx
kasbah: I read that advice too but I'm not sure it applies here. The card data is getting loaded lazily so shouldn't it be guaranteed to be "in view" when rendered? If so then why the additional lazy loading? We want to see the image now!

In practice it's likely imperceptible but we might as well make it make sense? #515 it looks more complicated, passing in the `priority` as a prop to the card. What do we gain by doing that?

kasbah: Hmmm... alright then.

frontend/src/components/ProjectCardGrid/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface ProjectCardGridProps {
+  projects: Array<Project>
+  intersectionObserverRef?: IntersectionObserverRef
+}
+
+const ProjectCardGrid = ({
+  projects,
+  intersectionObserverRef,
+}: ProjectCardGridProps) => {
+  // Set the marker to load more projects 6 cards bore the end of the grid.
```
kasbah, line 15:
```suggestion
  // Set the marker to load more projects to 6 cards before the end of the grid.
```

frontend/src/components/ProjectCardGrid/index.tsx
AbdulrhmnGhanem: Resolved in f592c2a.

kasbah (changes requested):

PR #512, opened 2022-11-24
Delete processor intermediate files

kasbah:
Removes all the temp files and those that get uploaded to s3. Just deleting everything at the end of the processing would have been easier but I decided to keep the removal of files as close as possible to the creation of the files. I think it makes more sense this way because it decouples the tasks from the rest of the app.

### DIFF
--- a/processor/src/tasks/processGerbers/index.ts
+++ b/processor/src/tasks/processGerbers/index.ts
@@ -223,9 +223,8 @@ export default async function processGerbers(
       job.updateProgress({ status: 'failed', file, error, outputDir })
     }
   }
-  await Promise.all(
-    filePaths.concat(plottedGerbers.gerbers).map(fp => fs.rm(fp, { force: true })),
-  )
+
+  await Promise.all(filePaths.map(fp => fs.rm(fp, { force: true })))
 }
 
 async function generateTopLargePng(topSvgPath, stackup, topLargePngPath) {
--- a/processor/src/tasks/processGerbers/index.ts
+++ b/processor/src/tasks/processGerbers/index.ts
@@ -1,7 +1,7 @@
 import globule from 'globule'
 import Jszip from 'jszip'
+import fs from 'node:fs/promises'
 import path from 'node:path'
-import { promises as fs } from 'node:fs'
 import { Job, JobData } from '../../job.js'
 import * as s3 from '../../s3.js'
 import { exec, execEscaped } from '../../utils.js'
@@ -223,6 +223,8 @@ export default async function processGerbers(
       job.updateProgress({ status: 'failed', file, error, outputDir })
     }
   }
+
+  await Promise.all(filePaths.map(fp => fs.rm(fp, { force: true })))
 }
 
 async function generateTopLargePng(topSvgPath, stackup, topLargePngPath) {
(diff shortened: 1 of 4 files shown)


processor/src/tasks/processGerbers/index.ts
```diff
@@ -223,6 +223,9 @@ export default async function processGerbers(
       job.updateProgress({ status: 'failed', file, error, outputDir })
     }
   }
+  await Promise.all(
+    filePaths.concat(plottedGerbers.gerbers).map(fp => fs.rm(fp, { force: true })),
+  )
```
kasbah, line 228:
```suggestion
```

processor/src/tasks/processGerbers/index.ts
```diff
@@ -1,7 +1,7 @@
 import globule from 'globule'
 import Jszip from 'jszip'
+import fs from 'node:fs/promises'
```
AbdulrhmnGhanem, line 3: Shouldn't these get deleted too? 

https://github.com/kitspace/kitspace-v2/blob/89a686c48bb4595f69bfd728a7e1bfd9bd739371/processor/src/tasks/processGerbers/index.ts#LL11C3-L11C3

AbdulrhmnGhanem: I think each job should delete its `outputFiles` when it's done.

processor/src/tasks/processKicadPCB/index.ts
```diff
@@ -1,15 +1,25 @@
...
+
 async function processKicadPCB(
   job: Job,
-  { inputDir, kitspaceYaml = {}, outputDir }: Partial<JobData>,
+  {
+    inputDir,
+    kitspaceYaml = {},
+    outputDir,
+    tmpDir,
+  }: ProcessKicadPcbData & Partial<JobData>,
```
AbdulrhmnGhanem, line 22: This type name is confusing

```suggestion
export type tmpDirPath = string
async function processKicadPCB(
  job: Job,
  {
    inputDir,
    kitspaceYaml = {},
    outputDir,
    tmpDir,
  }: tmpDirPath & Partial<JobData>,
```

processor/src/tasks/processGerbers/index.ts
kasbah: Oops, I accidentally removed the whole deletion. I've amended the last commit.

processor/src/tasks/processKicadPCB/index.ts
```diff
@@ -56,47 +66,47 @@ async function processKicadPCB(
...
 
     const relativeKicadPcbFile = path.relative(inputDir, kicadPcbFile)
     const inputFiles = { [relativeKicadPcbFile]: { type: 'kicad', side: null } }
+    await fs.rm(layoutSvgPath)
     return { inputFiles, gerbers }
   } catch (error) {
     for (const file of filePaths) {
       job.updateProgress({ status: 'failed', file, error, outputDir })
     }
+    await fs.rm(layoutSvgPath, { force: true })
```
AbdulrhmnGhanem, line 82: This should be in the `finally` block like the `processPCB`.

processor/src/tasks/processGerbers/index.ts
```diff
@@ -223,6 +223,8 @@ export default async function processGerbers(
       job.updateProgress({ status: 'failed', file, error, outputDir })
     }
   }
+
+  await Promise.all(filePaths.map(fp => fs.rm(fp, { force: true })))
```
AbdulrhmnGhanem, line 227: `top-meta.png` isn't in `filePaths`
image

processor/src/tasks/processKicadPCB/index.ts
kasbah: Changed in 179f9599d9dbbc1

processor/src/tasks/processGerbers/index.ts
kasbah: Fixed with 378da7b7c2b8195

AbdulrhmnGhanem (approved):

PR #515, opened 2022-11-26
Improve card thumbnail

### LINKED ISSUE #510: Use use higher res thumbnail when needed
On kitspace v1 we load `top-large.png` for thumbnails on mobile because the thumbnail is actually bigger. I'm not sure I ever really understood why that is (something to do with display density?) but right now on kitspace v2 the thumbnails get scaled up (and look slightly fuzzy) on my mobile. We should fix that. 

Maybe we should use imageset or let next/image do its magic here?

v1:
signal-2022-11-24-175209

v2:
signal-2022-11-24-175130_002

AbdulrhmnGhanem:
Fixes #510 

> Maybe we should use imageset or let next/image do its magic here?

`next/image` uses `srcset` so I used `top-large.png` and delegated optimization to next.

### DIFF
--- a/frontend/src/components/ProjectCard/index.tsx
+++ b/frontend/src/components/ProjectCard/index.tsx
@@ -12,6 +12,7 @@ interface ProjectCardProps {
   summary: string
   ownerName: string
   repoName: string
+  priority?: boolean
   intersectionObserverRef?: IntersectionObserverRef
 }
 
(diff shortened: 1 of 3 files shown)


kasbah:
Good idea, this seems to still be fetching `top.png` on review deploy though.

image

AbdulrhmnGhanem:
Where is this comment? I can't find it on GitHub.

kasbah:
I deleted a comment because I noticed it wasn't actually deployed to review.

PR #526, opened 2022-12-09
Create a script to import projects

### LINKED ISSUE #473: Script for migrating projects from kitspace.org
What could be a good workflow for migrating everything to alpha-2 kitspace-v2? Maybe a script that does something like: 

- Get https://raw.githubusercontent.com/kitspace/kitspace/master/boards.txt
- Create all the users required
- Add all the repos required
- Makes sure redirects exist for:
    - `/boards/*/user/project/*` to `/user/project/*`
    -  `/interactive_bom/?*/user/project` to `/user/project/IBOM` 
    - `/interactive_bom/?*/user/project/sub` to `/user/project/sub/IBOM` 
    - maybe we just add a permanent wildcard redirects in nginx? 
    - we will have to make sure with and without trailing slashes re-directs

AbdulrhmnGhanem:
- Create a script to import v1 projects
- CI: add action to copy the migration script to our hosts
- Add repo name to zod error message
- Expose gitea port to host in deployment environment

##
Closes #473.


AbdulrhmnGhanem:
For registry boards, I think we have to use a forked version, I couldn't find a way to change the `HEAD` from the API, and even if there was a way, it will break mirroring.

.github/workflows/copy-import-script.yml
```diff
@@ -0,0 +1,50 @@
...
+          target: '/home/deploy'
+
+      - name: Make the script executable on hosts
+        uses: appleboy/ssh-action@master
+        with:
+          host: ${{ matrix.host }}
+          username: deploy
+          key: ${{ secrets.STAGING_SSH_PRIVATE_KEY }}
+          script: |
+            cd /home/deploy && mv target/importBoardsTxt importBoardsTxt && chmod +x importBoardsTxt && rm -rf target
```
kasbah, line 50: Why are you copying a whole directory, moving a file out of it and then deleting it? 

Also feel like we don't need special actions for this, could just use `scp` and `ssh` but since this script is a stop-gap anyway I'm willing to ignore it.

kasbah (changes requested):

.github/workflows/copy-import-script.yml
AbdulrhmnGhanem: > Why are you copying a whole directory, moving a file out of it and then deleting it?

The `appleboy/scp-action` action can't copy files, only dirs.

> could just use scp and ssh 

I did so during the development, but I felt we can miss updating the script on one of the servers if we made a change to it. This way we are sure we have the latest script on all the staging severs.

kasbah: I meant, you can just use `scp` and `ssh` directly instead of relying on someone else's action. Especially if it has weird limitations like that.

scripts/create_gitea_admin_token.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# Create a Gitea admin user in gitea and generate an API token for it.
+# Usage: create_gitea_admin_user <username> <password> <email>
+# Example: create_gitea_admin_user admin admin
+if [ $# -ne 3 ]; then
+    echo "Usage: create_gitea_admin_user <username> <password> <email>"
```
kasbah, line 6:
```suggestion
    echo "Usage: $(basename "$0") <username> <password> <email>"
```

scripts/README.MD
```diff
@@ -0,0 +1,33 @@
...
+      --giteaUrl: Gitea URL (default: http://localhost:3333)
+      --adminToken: Gitea admin API token (classic)
+      --githubToken: GitHub API token
+      --numberOfRepos: Number of repositories to import (default: 1000)
+      --help: Show this help
+```
+### To generate `adminToken` run `create_gitea_admin_token.sh`
+```console
+$ ./scripts/create_gitea_admin_token.sh
+Usage: create_gitea_admin_user <username> <password> <email>
```
kasbah, line 16:
```suggestion
Usage: create_gitea_admin_token.sh <username> <password> <email>
```

kasbah (commented):
I tried this briefly, seems to work well, though it does run out of space on the server. I think #543 should be prioritized since merging this without that would break staging when we run it. 

Is there a reason that we copy over `importBoardsTxt` but not `create_gitea_admin_token.sh`? Seems I can't run it without creating a token so I had to copy it over to the server myself. Would be extra nice if the script just handled all the token creation/re-use when I run `./importBoardsTxt`. Maybe we could also embed a Github API token into the script when it's built on the CI?

AbdulrhmnGhanem:
> Maybe we could also embed a Github API token into the script when it's built on the CI?

I think this is doable https://github.com/denoland/deno/issues/9152#issuecomment-773522311.

> Would be extra nice if the script just handled all the token creation/re-use when I run ./importBoardsTxt.

OK, I will make it do so.

scripts/README.md
```diff
@@ -0,0 +1,27 @@
...
+Usage: importBoardsTxt [options]
+    options:
+      --giteaUrl: Gitea URL (default: http://localhost:3333)
+      --adminToken: Gitea admin API token (default: generated automatically)
+      --githubToken: GitHub API token (classic) (Embedded into the script in staging servers.)
+      --numberOfRepos: Number of repositories to import (default: 1000)
+      --tokenOnly: Only generate the admin token and exit.
+      --help: Show this help
+```
+### To generate `adminToken` run `create_gitea_admin_token.sh`
```
kasbah, line 14:
```suggestion
### Generating `adminToken`
```

kasbah (changes requested):
Tested it out, worked great! Some minor fixes remaining.

.github/workflows/copy-import-script.yml
```diff
@@ -0,0 +1,50 @@
...
+    environment: staging
+    strategy:
+      matrix:
+        host:
+          [
+            abdo-dev.staging.kitspace.dev,
+            kaspar-dev.staging.kitspace.dev,
+            review.staging.kitspace.dev,
+            master.staging.kitspace.dev,
+          ]
```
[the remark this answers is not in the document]
AbdulrhmnGhanem: I ran it, no changes!

PR #542, opened 2023-01-05
Add reachability healthcheck for web services

AbdulrhmnGhanem:
this healthcheck tests all the web services' reachability from the docker
network `http://<service>` and for external server access
`https://<service>.$KITSPACE_DOMAIN` in prod/staging envs.
If one of the services is unreachable it will reload Nginx. If reloading
nginx fails it exits and marks the container as unhealthy.
The healthcheck only doesn't restart the container if it's unhealthy
though. Restarting unhealthy containers should be done by the host
machine.


AbdulrhmnGhanem:
And I don't think using the same port in the containers is related, the ports aren't bound to host ports. Maybe it's because the ports aren't exposed? 

image

nginx/reachability_healthcheck
```diff
@@ -0,0 +1,29 @@
...
+# network (`http://<service>`) and for external server access
+# (`https://<service>.$KITSPACE_DOMAIN`) in production and staging envrionments.
+# If one of the services is unreachable it will reload nginx. If reloading
+# nginx fails it exits marks the container as unhealthy.
+
+reload_nginx() {
+    echo "Reloading nginx!"
+    # We try to reload nginx first, if this fails we mark the container as unhealthy
+    nginx -s reload || exit 1
+    # Sleep for 10 seconds to give nginx time to reload before we check hte next service.
```
kasbah, line 15:
```suggestion
    # Sleep for 10 seconds to give nginx time to reload before we check the next service.
```

nginx/reachability_healthcheck
```diff
@@ -0,0 +1,51 @@
...
+        test_reachability "https://meilisearch.$KITSPACE_DOMAIN" "Meilisearch" || reload_nginx "meilisearch external"
+    fi
+}
+
+reload_nginx() {
+    echo "$1 is unreachable!"
+    echo "Reloading nginx!"
+    # We try to reload nginx first, if this fails we mark the container as unhealthy
+    nginx -s reload || report_healthcheck_failure "$1"
+}
```
kasbah, line 30: So this won't report anything if reloading solves it?

docker-compose.yml
```diff
@@ -52,13 +52,23 @@ services:
...
       - meilisearch
     ports:
       - '${KITSPACE_DEV_PORT}:80'
+    healthcheck:
+      test: ["CMD", "reachability_healthcheck"]
+      interval: 1m
+      timeout: 1m
+      # If any of the services is unreachable for 4 minutes, mark the container as unhealthy.
+      retries: 4
+      start_period: 2m
```
kasbah, line 71: TIL

kasbah (changes requested):
Is this working nicely now then?

> I think we should configure the notifications to only send emails if the issue is in staging environment and it happend n times in a certain interval.

Have you done that?

nginx/reachability_healthcheck
```diff
@@ -0,0 +1,51 @@
...
+# If one of the services is unreachable it will reload nginx. If reloading
+# nginx fails it exits marks the container as unhealthy.
+
+healthcheck() {
+    # Make sure the hitting domain name returns kitspace.org and .gitea subdomain returns gitea.kitspace.org
+    test_reachability "http://frontend:3000" "Kitspace" || reload_nginx "frontend"
+    test_reachability "http://gitea:3000" "Gitea"  || reload_nginx "gitea"
+    test_reachability "http://meilisearch:7700" "Meilisearch" || reload_nginx "meilisearch"
+
+    # Run the prevoious two checks for the domain name if it's not kitspace.test
```
kasbah, line 17:
```suggestion
    # Run the previous two checks for the domain name if it's not kitspace.test
```

nginx/reachability_healthcheck
AbdulrhmnGhanem: Yes!

kasbah: Would be nice if it was still recorded.

AbdulrhmnGhanem:
> Have you done that?

I take this suggestion back. I thought this issue would get reported frequently. But we have had it deployed for two weeks now, we didn't have any false positives so I will leave it as is.

nginx/reachability_healthcheck
AbdulrhmnGhanem: In daf9fbfa, I made it report the failure if a service is unreachable, then it tries reloading Nginx if reloading fails it reports this too.
Also, I removed the docker network checks because I verified it isn't an issue with the docker network by getting a shell inside the Nginx container and `curl`ing the container: the result was OK but the external domains were mixed.

PR #550, opened 2023-01-13
Update docs for clearing s3 buckets

AbdulrhmnGhanem:
Add `--recursive` flag to make sure that all files are actually cleared.

### DIFF
--- a/processor/README.md
+++ b/processor/README.md
@@ -49,5 +49,5 @@ Minio Docker container. To clear out the data SSH into the staging server and
 remove the bucket. E.g.:
 
 ```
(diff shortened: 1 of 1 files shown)


kasbah:
Oh, but how can files still be there if the bucket is removed?

AbdulrhmnGhanem:
> Single Local File and S3 Object Operations
       Some commands perform operations only on single files and  S3  objects.
       The following commands are single file/object operations if no --recur-
       sive flag is provided.
           o cp
           o mv
           o rm

The bucket gets deleted by `rb` command, not `rm`. Also, it has to be empty before deleting it.
I tried it but I got a permission error.  Perhaps the access key in staging servers doesn't have this privilege.

PR #551, opened 2023-01-13
Don't cache `useProjectAsset` head request

### LINKED ISSUE #507: S3 CORS errors in staging
https://review.staging.kitspace.dev/tester/test_tube_photometer/tt_nano_HAT_b1

image

image

AbdulrhmnGhanem:
Fixes #507.
It won't cause any performance issues; the body size is zero.

### DIFF
--- a/frontend/src/hooks/useKitspaceYAML.js
+++ b/frontend/src/hooks/useKitspaceYAML.js
@@ -1,7 +1,7 @@
 import useSWR from 'swr'
 import getConfig from 'next/config'
 
-const fetcher = (...args) => fetch(...args).then(r => r.json())
+import { fetcher } from './s3HeadFetcher'
 
 const assetUrl = getConfig().publicRuntimeConfig.KITSPACE_ASSET_URL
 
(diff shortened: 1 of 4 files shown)


kasbah:
Isn't `useThumbnail` (and maybe other hooks that do `HEAD` on s3) also affected by this?

frontend/src/hooks/useKitspaceYAML.js
```diff
@@ -1,7 +1,7 @@
 import useSWR from 'swr'
 import getConfig from 'next/config'
 
-const fetcher = (...args) => fetch(...args).then(r => r.json())
+import { fetcher } from './_s3HeadFetcher'
```
kasbah, line 4: What's with the leading underscore?

AbdulrhmnGhanem: I want to keep the file in the `hooks` dir, but it isn't a hook. So I am trying to signal this.

kasbah: Let's not start doing that. Feels like you are working around the issue that having a "hooks" folder doesn't make sense. I'm sure it's a matter of taste but I'd rather the code base was organized around the features of our app and less around React or Next internals. Once we reach alpha-2 it would probably be a good time to re-organize it.

AbdulrhmnGhanem: Removed the dash.

PR #552, opened 2023-01-13
Don't treat emails as relative URLs in the README

### LINKED ISSUE #454: Emails are treated as relative URLs in readme
https://github.com/kelu124/bomanz

|v1|v2|
|---|---|
|image|image|

AbdulrhmnGhanem:
Fixes #454

### DIFF
--- a/processor/src/utils.ts
+++ b/processor/src/utils.ts
@@ -59,7 +59,9 @@ export function findKicadSchematic(inputDir, files, kitspaceYaml) {
 }
 
 export function isRelativeUrl(uri: string) {
-  return !uri.match(/^https?:\/\//)
+  // We can't use `[a-z0-9]*:` for matching mailto, and others, because
+  // a `:` in the path part of the url will also match.
+  return !uri.match(/^[a-z0-9]*?:\/\/|^(mailto|tel|sms|magnet):/)
 }
 
 export function normalizeRelativeUrl(url: string, rootFolder: string) {
--- a/processor/test/unit/readme_processing.test.ts
+++ b/processor/test/unit/readme_processing.test.ts
@@ -1,14 +1,13 @@
-import { describe, it } from 'vitest'
+import { describe, expect, it } from 'vitest'
 import { renderMarkdown } from '../../src/tasks/processReadme/renderMarkdown'
-
-const invalidTagReadme = `# 8BitmixtapeNEO_ShenzhenReady
-
-summary: <a summary for your project>
-site: https://8bitmixtape.github.io/NeoWiki
-`
 
 describe('readme processing', function () {
   it('processes readme with invalid <a> tags', async function () {
+    const invalidTagReadme = `
+    # 8BitmixtapeNEO_ShenzhenReady
+    summary: <a summary for your project>
+    site: https://8bitmixtape.github.io/NeoWiki
+    `
     const args = {
       rawMarkdown: invalidTagReadme,
       ownerName: 'test',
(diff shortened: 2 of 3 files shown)


processor/src/utils.ts
```diff
@@ -59,7 +59,7 @@ export function findKicadSchematic(inputDir, files, kitspaceYaml) {
 }
 
 export function isRelativeUrl(uri: string) {
-  return !uri.match(/^https?:\/\//)
+  return !uri.match(/^https?:\/\/|^mailto:/)
```
kasbah, line 62: I think it makes sense to make this generic over the protocol. We might get the same problem with e.g. `ssh://` links or something else.

AbdulrhmnGhanem: I found that the sanitizer only allows `http(s)` and `mailto`. I changed that :point_down:
https://github.com/kitspace/kitspace-v2/blob/f51a0792074fdbb19caaecccb1e5a68ebfb4907c/processor/src/tasks/processReadme/renderMarkdown.ts#L30

And changed the `isRelativeUrl` accordingly.

Do we need to allow any other protocols?

kasbah: No, we don't need them. It's fine and in fact better if the others don't get "linkified" at all.

kasbah: Actually, wait, maybe it handy to all of these linkified? I don't think they are used very much but might as well?

AbdulrhmnGhanem: They are already being linkified, no extra work is required. The only thing I did is stop treating them as relative URLs.

kasbah: But before the sanitizer would strip them out completely?

processor/test/unit/readme_processing.test.ts
```diff
@@ -20,4 +19,34 @@ describe('readme processing', function () {
...
+      originalUrl: 'https://github.com/kelu124/bomanz/',
+      defaultBranch: 'main',
+    }
+    const html = await renderMarkdown(args)
+    expect(html).toContain('href="mailto:kelu124@gmail.com"')
+  })
+
+  it('does not treat urls starting with a TCP protocol as a relative URL', async function () {
+    const readmeWitSshURL =
+      "I don't know why someone would put ssh url in a readme: [but just in case](ssh://git@github.com:kitspace/kitspace-v2.git)"
```
kasbah, line 40: ^_^

processor/src/utils.ts
AbdulrhmnGhanem: Yes!

processor/test/unit/readme_processing.test.ts
```diff
@@ -20,4 +19,34 @@ describe('readme processing', function () {
     // if it throws an error this test will fail
     await renderMarkdown(args)
   })
+
+  it('does not treat mailto as a relative URL', async function () {
+    const readmeWithMailto = 'Copyright Kelu124 (kelu124@gmail.com ) 2015-2018'
```
kasbah, line 24:
```suggestion
    const readmeWithMailto = 'Copyright Example (example124@example.com ) 2015-2018'
```

processor/test/unit/readme_processing.test.ts
```diff
@@ -20,4 +19,34 @@ describe('readme processing', function () {
...
+    const args = {
+      rawMarkdown: readmeWithMailto,
+      ownerName: 'test',
+      repoName: 'test',
+      readmeFolder: 'test',
+      originalUrl: 'https://github.com/kelu124/bomanz/',
+      defaultBranch: 'main',
+    }
+    const html = await renderMarkdown(args)
+    expect(html).toContain('href="mailto:kelu124@gmail.com"')
```
kasbah, line 35:
```suggestion
    expect(html).toContain('href="mailto:example124@example.com"')
```

PR #561, opened 2023-02-16
Delete user UI functionalities

AbdulrhmnGhanem:
- Disable sign in/up
- Make `Add a project` button redirect to instructions in the repo
- Delete `/projects/new` page
- CI: disable cypresss

All tests depend on importing repos, we have disabled importing repos
from the UI. We have to write a command to import the repos without the
UI and re-enable the tests because it would be a bit dangerous to go
live without tests.


AbdulrhmnGhanem:
The tests are [passing](https://github.com/kitspace/kitspace-v2/actions/runs/4260342709) on the `abdo-dev` branch but are failing here because it runs an older workflow file (the pull-request target thing).

kasbah (changes requested):
Could this be split into the "deleting project upload" and "deleting the sign up" so we can revert this individually when we want that functionality?

AbdulrhmnGhanem:
> Could this be split into the "deleting project upload" and "deleting the sign up" so we can revert this individually when we want that functionality?

I can do it but I think we shouldn't squash this PR; every commit is a time step we can get back to.

AbdulrhmnGhanem:
I rolled back the default value for the env variable, the script won't GITHUB_TOKEN must be present for the script to work as expected in CI. So I think we need to move the `GITHUB_TOKEN` secret from the deployment environment to global repo secrets (This a read-only token required for cloning the repos so it's not so dangerous).

kasbah:
Shouldn't removing the tests and removing functionality be in the same commit though? I'd like to make reverting these as easy as possible...

kasbah:
> I rolled back the default value for the env variable, the script won't GITHUB_TOKEN must be present for the script to work as expected in CI. So I think we need to move the `GITHUB_TOKEN` secret from the deployment environment to global repo secrets (This a read-only token required for cloning the repos so it's not so dangerous).

Sorry, I'm not sure what you mean. The `GITHUB_TOKEN` is set automatically in Github actions, it's not in our environment secrets.

AbdulrhmnGhanem:
> Shouldn't removing the e2e tests and removing functionality be in the same commit though?

Done

> Sorry, I'm not sure what you mean. The GITHUB_TOKEN is set automatically in Github actions, it's not in our environment secrets

We need to pass the step's env, which is done in this pull request: https://github.com/kitspace/kitspace-v2/pull/561/files#diff-3f5366f6d6df3ec1179e5efadc6f350bfa88eebf4e2da589b4d94ccb85ae5e94R330-R335. However, since `pull_request_target` runs the file in master, the `env` isn't being set as required for the script to work. Sorry for any confusion caused earlier, I was a bit sleepless when I replied! :smile:

scripts/pre_e2e.sh
```diff
@@ -6,3 +6,18 @@ set -Eeuo pipefail
...
+
+if [[ -z $GITHUB_TOKEN ]]; then
+    # The tests are running locally.
+    docker-compose up -d
+else
+    docker-compose -f docker-compose.yml -f docker-compose.deploy.yml up -d
+    until docker logs kitspace_gitea_1 | grep 'ORM engine initialization successful' ; do sleep 3s; done
+fi
+
+token="$(deno run --allow-env --allow-net --allow-run ./scripts/importBoardsTxt.ts --tokenOnly)"
```
kasbah, line 20: It doesn't seem like you need the GITHUB_TOKEN to do any of this?

kasbah:
I've pushed some fixup commits to this branch. If you agree with them then rebase and squash them in.

.github/workflows/docker.yml
```diff
@@ -344,6 +356,7 @@ jobs:
           COMMIT_INFO_SHA: '${{ github.event.head_commit.id }}'
           COMMIT_INFO_TIMESTAMP: '${{ github.event.head_commit.timestamp }}'
           COMMIT_INFO_BRANCH: '${{ github.ref_name || github.head_ref }}'
+          CYPRESS_ADMIN_TOKEN: ${{ env.CYPRESS_ADMIN_TOKEN }}
```
kasbah, line 359:
```suggestion
```
I don't think you need to do this. It's in the env already.

e2e/cypress/support/commands.js
```diff
@@ -59,3 +61,104 @@ Cypress.Commands.add('forceVisit', (path, timeout) => {
...
+  cy.waitForStatusCode(
+    `http://gitea.kitspace.test:3000/${user.username}/${repoName}`,
+  )
+})
+
+Cypress.Commands.add('mirrorRepo', (remoteRepo, repoName, user) => {
+  const endpoint = 'http://gitea.kitspace.test:3000/api/v1/repos/migrate'
+  const headers = {
+    'Content-Type': 'application/json',
+    Authorization: `token ${Cypress.env('CYPRESS_GITEA_ADMIN_TOKEN')}`,
```
kasbah, line 78:
```suggestion
    Authorization: `token ${Cypress.env('GITEA_ADMIN_TOKEN')}`,
```

e2e/cypress/support/commands.js
```diff
@@ -59,3 +61,104 @@ Cypress.Commands.add('forceVisit', (path, timeout) => {
...
+    if (!response.status === 201) {
+      throw new Error('Failed to mirror repo')
+    }
+  })
+})
+
+Cypress.Commands.add('createGiteaUser', user => {
+  const headers = {
+    'Content-Type': 'application/json',
+    Authorization: `token ${Cypress.env('CYPRESS_GITEA_ADMIN_TOKEN')}`,
```
kasbah, line 110:
```suggestion
    Authorization: `token ${Cypress.env('GITEA_ADMIN_TOKEN')}`,
```

kasbah:
The pre_e2e.sh thing is annoying and doesn't export env vars  as I expected. Can we please just get the admin token in cypress when we run it?

e2e/cypress/support/commands.js
AbdulrhmnGhanem: It has to start with `cypress_/CYPRESS_` to be read by cypress automatically  https://docs.cypress.io/guides/guides/environment-variables#Option-3-CYPRESS_

e2e/cypress/support/commands.js
kasbah: Yes but when you retrieve it the `CYPRESS_` prefix is omitted 

> export CYPRESS_HOST=laura.dev.local
> ...
> 
> Will yield these results inside a test file:
> 
>> Cypress.env('HOST') // 'laura.dev.local'

AbdulrhmnGhanem:
> The pre_e2e.sh thing is annoying and doesn't export env vars as I expected. Can we please just get the admin token in Cypress when we run it?

As I have explained before we can only generate the token from the host machine. So I update the README with streamlined instructions to run the tests locally, for example, to open the GUI:

```console
yarn --cwd e2e gui --env GITEA_ADMIN_TOKEN=$(scripts/pre_e2e.sh)
```

scripts/pre_e2e.sh
```diff
@@ -4,5 +4,18 @@ set -Eeuo pipefail
 # Generate big files which are used during testing
 # `uploadProject.spec.js` -> should reject files bigger than `MAX_FILE_SIZE`
 auto_gen_fixtures_path=e2e/cypress/fixtures/auto-gen
-mkdir $auto_gen_fixtures_path
-truncate -s "${MAX_FILE_SIZE}" "${auto_gen_fixtures_path}/big.txt"
+rm -rf $auto_gen_fixtures_path && mkdir $auto_gen_fixtures_path
+truncate -s "${MAX_FILE_SIZE:-6M}" "${auto_gen_fixtures_path}/big.txt"
```
kasbah, line 8: I would prefer to error if this is not set.

scripts/pre_e2e.sh
```diff
@@ -4,5 +4,18 @@ set -Eeuo pipefail
...
+    docker-compose -f docker-compose.yml -f docker-compose.deploy.yml up -d
+    until docker logs kitspace_gitea_1 | grep -q 'ORM engine initialization successful' ; do sleep 3s; done
+else
+    # The tests are running locally.
+    docker-compose up -d
+fi
+
+token="$(deno run --allow-env --allow-net --allow-run ./scripts/importBoardsTxt.ts --tokenOnly)"
+# reutrn the token so it can be used as $(./scripts/pre_e2e.sh)
+echo $token
```
kasbah, line 21: I don't think giving this script two purposes is a good idea.

AbdulrhmnGhanem: I think of it as the stuff we need to do before running e2e (the build step), and like any other build step, it can involve multiple steps.

scripts/pre_e2e.sh
AbdulrhmnGhanem: Resolved in c8d2b99.

scripts/pre_e2e.sh
kasbah: I guess but if the instructions say `yarn --cwd e2e gui --env GITEA_ADMIN_TOKEN=$(scripts/pre_e2e.sh)` then the fact that it also creates a test-data file when you do that is surprising behaviour. I suggest just saying:

```
GITEA_ADMIN_TOKEN="$(deno run --allow-env --allow-net --allow-run ./scripts/importBoardsTxt.ts --tokenOnly)"
yarn --cwd e2e gui --env GITEA_ADMIN_TOKEN="${GITEA_ADMIN_TOKEN}"
```
(shouldn't it by `CYPRESS_GITEA_ADMIN_TOKEN` anyway?)

Ideally `pre_e2e.sh` should be called `generate_test_fixture.sh` or similar so it's less tempting to make it a dumping ground. Also `importBoardsTxt.ts --tokenonly` should be a separate script (`createGiteaAdminToken.ts`) but I'm willing to let these go in the interest of shipping. 

In general you should think twice if you are making any script/function/class/whatever do two things because it makes what it's going to do much less clear.

README.md
```diff
@@ -335,22 +335,39 @@ We configure our staging servers using [Ansible](https://docs.ansible.com/ansibl
 ```console
 docker-compose -f docker-compose.yml -f docker-compose.override.yml -f docker-compose.e2e.yml up e2e
 ```
+##
+
+> 🚧 **GITEA_ADMIN_TOKEN is needed because we have hid gitea and it's inaccessible to other containers!**
```
kasbah, line 340:
```suggestion
> Note: The GITEA_ADMIN_TOKEN is needed because the Gitea service is hidden and inaccessible to other containers!
```

scripts/pre_e2e.sh
AbdulrhmnGhanem: Resolved in 0872bab.

PR #563, opened 2023-02-16
Hide gitea

### LINKED ISSUE #474: Hiding Gitea
Hide away `gitea.` sub-domain except for the endpoints we use from the client side.

AbdulrhmnGhanem:
Proxy api, and raw content requests to gitea, otherwise return 404.
The raw content is for readme images.

Closes #474.

### DIFF
--- a/nginx/kitspace.template
+++ b/nginx/kitspace.template
@@ -4,7 +4,7 @@ server {
 
     server_name @KITSPACE_DOMAIN;
 
-    location /robots.txt {
+    location = /robots.txt {
         add_header Content-Type text/plain;
         return 200 "@KITSPACE_ROBOTS_TXT";
     }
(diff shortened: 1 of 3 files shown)


nginx/kitspace.template
```diff
@@ -59,6 +61,10 @@ server {
         }
         proxy_pass http://gitea:3000;
     }
+
+    location ~ / {
+        return 404;
+    }
```
kasbah, line 67: This seems to be taking precedence over /robots.txt. Maybe just remove the robots.txt location in the gitea server? If most things are 404 anyway then it's probably not needed.

AbdulrhmnGhanem: Resolved in 29820971.

PR #564, opened 2023-02-17
Fix history navigation for search

AbdulrhmnGhanem:
Before this commit using the browser history navigation didn't work for
the search. i.e, if you type a `q1` in the search bar, and submit, then type
`q2` and submit, if you go back using the browser button, the page would
still show the results of `q2` instead of `q1`

### DIFF
--- a/frontend/src/contexts/SearchContext.tsx
+++ b/frontend/src/contexts/SearchContext.tsx
@@ -11,6 +11,7 @@ const SearchProvider = ({ children, initialQuery }: SearchProviderProps) => {
 
   const {
     query: { q },
+    events,
   } = useRouter()
 
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


frontend/src/contexts/SearchContext.tsx
```diff
@@ -19,6 +20,18 @@ const SearchProvider = ({ children, initialQuery }: SearchProviderProps) => {
...
   }, [q])
 
+  events.on('routeChangeComplete', (url: string) => {
+    /*
+     * Handle history navigation by update the search query in the context when the url changes.
+     * We read the query parameter from the url instead of using the router query
+     * because the router query gets updated asynchronously.
+     */
+    const latestQuery = new URLSearchParams(url.split('?')[1]).get('q')
+    if (latestQuery !== null) {
```
kasbah, line 30:
```suggestion
    if (latestQuery != null) {
```

If there is no "?" in the `url` then latestQuery will be `undefined` which `!== null`. Generally when checking for null/undefined I just use the lax equality check instead of the stricter triple one.

PR #566, opened 2023-02-24
Port `/newsletter` from v1

AbdulrhmnGhanem:
The subscription submission hasn't been handled yet.


kasbah:
I believe the buttondown API changed and we no longer need a netlify function for this.

PR #570, opened 2023-04-06
Upgrade to KiCad 7

kasbah:
- Upgrade our scripts to KiCad 7
- Upgrade IBOM for compatibility, IBOM has native Eagle support now so we no longer need to convert those to KiCad

### DIFF
--- a/processor/Dockerfile
+++ b/processor/Dockerfile
@@ -3,6 +3,7 @@ FROM ubuntu:22.04
 ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update && apt-get install -qq -y software-properties-common curl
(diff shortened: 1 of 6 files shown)


kasbah:
Any improvements on this from porting it to EDeA?

AbdulrhmnGhanem:
I added styles and made XML parsing safer, do you want to port that here?

kasbah:
Styles, I think not because I'd like to set them from the outside so they can be switched out.

Safer, sure, but it's not a priority since the SVG/XML is generated on our servers from KiCad.

Guess is good to merge then?

AbdulrhmnGhanem:
Did you check that IBOM is working as expected on importing a new project?  if yes, then it's OK to merge.

PR #575, opened 2023-05-06
Make `copy-import-script` workflow branch specific

AbdulrhmnGhanem:


.github/workflows/copy-import-script.yml
```diff
@@ -31,19 +23,23 @@ jobs:
...
-      - name: Copy the script to hosts
+      - name: Extract branch name
+        run: echo "name=$(echo ${GITHUB_REF#refs/heads/})" >>$GITHUB_OUTPUT
+        id: branch_name
+
+      - name: Copy the script to host
         uses: appleboy/scp-action@master
         with:
-          host: ${{ matrix.host }}
+          host: ${{ steps.branch_name.outputs.name }}.staging.kitspace.dev
```
kasbah, line 33: There seems to be a `${{ github.ref_name }}` now which we should probably start using as it's a bit simpler.

PR #584, opened 2023-05-21
Make logo.svg an import

kasbah:
This way it gets a hash appended and put in `_next/static` and so we can set the caching headers with all other static immutable assets.


AbdulrhmnGhanem:
It broke the logo alignment
|before|after|
|---|---|
|image|image|

frontend/src/components/NavBar/index.module.scss
```diff
@@ -53,11 +53,6 @@
 .logoImg {
   height: 46px;
   width: 160px;
```
kasbah, line 55:
```suggestion
  width: 160px;
  display: block;
```

kasbah:
I think it's fixed with https://github.com/kitspace/kitspace-v2/pull/584/commits/8bf7013edc50f26585335a90b6e39282b1f015ac.

PR #629, opened 2024-01-19
Switch to go-gitea docker image

AbdulrhmnGhanem:
Drop d1add0a, when `partinfo` is back

### DIFF
--- a/docker-compose.deploy.yml
+++ b/docker-compose.deploy.yml
@@ -8,7 +8,6 @@ services:
     restart: always
 
   gitea:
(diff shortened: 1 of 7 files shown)


AbdulrhmnGhanem:
I ran 
```sh
docker stop $(docker ps -a -q)
docker system prune --all --volumes
docker volume prune --filter all=1
aws s3 rm s3://kitspace-staging-abdo-dev --recursive
./importBoardsTxt
````
everything is working as expected.

PR #675, opened 2024-10-05
Download BoM files

AbdulrhmnGhanem:
1. Use https://github.com/kasbah/push-on-hold-off as a test fixture.
2. Migrate all the jsx files I touched to tsx
3. Add Plausible types
4. Download bom csv file optimizied for each retailer
5. Extract some inline styles into scss files

(unit tests for the new download feature and digikey's direct link)

##
## templates
1. Digikey remains a link
7. [Mouser.xls](https://github.com/user-attachments/files/17264725/BomTemplate.xls)
8. [RS.xls](https://github.com/user-attachments/files/17264730/RS-quote-template-xa.xls)
9. [Farnell.xlsx](https://github.com/user-attachments/files/17264732/Sample_Template_en_GB.xlsx)
10. Newark has the same file format as Farnell


kasbah:
Stick to `.csv` if you can. We don't want to load an XLSX library client side. Here is [LCSC_BomTemplate.csv](https://github.com/user-attachments/files/17266296/LCSC_BomTemplate.csv).

AbdulrhmnGhanem:
Sure, all of them accept csv but the templates are xlsx

kasbah:
One thing to watch out for is the Unicode encoding for the downloaded `.csv`. 

- Javascript might default to UTF-16, I can't remember. 
- Excel defaults to UTF-16 I'm pretty sure. 
- Does the browser do different things on different platforms? 

Should probably test downloading and uploading to vendor some BOMs with unicode characters on different platforms.

AbdulrhmnGhanem:
I will do it as CSV first and test. If I find it too hard to stratify every vendor, I will switch to XLSX.

AbdulrhmnGhanem:
The RS template is out of date, I downloaded the real BoM for a project and made the implementation according to it [cart.csv](https://github.com/user-attachments/files/17285788/cart.csv)

AbdulrhmnGhanem:
Is there a board that has the six retailers in the BoM? I want to use it as a test fixture.

kasbah:
Maybe  not, I can only think of https://kitspace.org/boards/github.com/kasbah/push-on-hold-off/ which has 5. If you send a PR for LCSC I'll merge it though :)

AbdulrhmnGhanem:
>  If you send a PR for LCSC I'll merge it though :)

https://github.com/kasbah/push-on-hold-off/pull/7

AbdulrhmnGhanem:
commit 7708f29

|before|after|
|---|---|
|image|image|

frontend/src/components/Board/BuyParts/Bom.tsx
```diff
@@ -44,13 +43,13 @@ const ExpandBom = ({ diff, collapsed, setCollapsed }) => {
     ) : null
 
   return (
-    <div style={{ paddingLeft: 1, paddingRight: 1 }}>
+    <div className={`${styles.expandWrapper}`}>
```
kasbah, line 46:
```suggestion
    <div className={styles.expandWrapper}>
```

frontend/src/components/Board/BuyParts/index.module.scss
```diff
@@ -1,8 +1,10 @@
 @import '../../breakpoint.scss';
 
 .notSelectable {
+  -webkit-user-select: none;
   user-select: none;
   cursor: default;
+  margin-left: 5px;
```
kasbah, line 7: Is the -webkit prefix one needed?

Why are we mixing margin concerns with notSelectable?

kasbah (changes requested):
The buttons are always green now when before they would be red when not all parts are in the BOM for that store.

frontend/src/components/Board/BuyParts/index.tsx
```diff
@@ -0,0 +1,464 @@
...
+function downloadBomCSV({
+  name,
+  lines,
+  multiplier,
+  buyAddPercent,
+  retailer,
+}: Order) {
+  const csvContent = `${csvBom(lines, multiplier, buyAddPercent, retailer)
+    .map((e: Array<string>) => e.join(','))
+    .join('\n')}\n`
```
kasbah, line 397: - Strings with commas in them need to be quoted
- Strings with quotes in them need to be escaped

frontend/src/components/Board/BuyParts/index.tsx
```diff
@@ -408,6 +408,20 @@ function redirectToStore(retailer: Retailer) {
...
+/**
+  Escape a CSV value according to RFC 4180.
+  https://www.ietf.org/rfc/rfc4180.txt
+ */
+function escapeCsvValue(value: string | number | undefined) {
+  if (
+    typeof value === 'string' &&
+    (value.includes(',') || value.includes('"') || value.includes('\n'))
+  ) {
+    return `"${value.replace(/"/g, '""')}"`
```
kasbah, line 420: Is that right, shouldn't it be `\"`?

frontend/src/components/Board/BuyParts/index.tsx
```diff
@@ -0,0 +1,479 @@
...
+
+function downloadBomCSV({
+  name,
+  lines,
+  multiplier,
+  buyAddPercent,
+  retailer,
+}: Order) {
+  const csvContent = `${csvBom(lines, multiplier, buyAddPercent, retailer)
+    .map((e: Array<CsVValue>) => e.map(escapeCsvValue).join(','))
```
kasbah, line 396:
```suggestion
    .map((e: Array<CsvValue>) => e.map(escapeCsvValue).join(','))
```

frontend/src/components/Board/BuyParts/index.tsx
```diff
@@ -0,0 +1,479 @@
...
+  totalLines: number
+  numberOfParts: number
+  icons: 'link' | 'download'
+}
+
+interface StoreIconProps {
+  retailer: string
+}
+
+type CsVValue = string | number | undefined
```
kasbah, line 459:
```suggestion
type CsvValue = string | number | undefined
```

frontend/src/components/Board/BuyParts/index.tsx
AbdulrhmnGhanem: I implemented it according to the RFC linked in the function docstring

image

kasbah (approved):

PR #679, opened 2024-11-22
Change buy parts label icons to "link" or "download"

kasbah:
PR for your branch @AbdulrhmnGhanem. Changing the button icons to:

image


AbdulrhmnGhanem:
isn't the Digikey icon a bit cluttered?

kasbah:
I guess, but it does express what's going to happen when you click it quite well. I don't have a better solution right now, removing any of the 3 icons just seems less clear.

PR #682, opened 2024-11-22
Use references in csv customer part numbers

kasbah:


kasbah:
This is also against your branch @AbdulrhmnGhanem

Issue #678, opened 2024-11-22
Unicode encoding issues in BOMs


kasbah:
![image](https://github.com/user-attachments/assets/94d44103-a43e-4448-8df1-24df58e49b4d)

Should be Würth. Appears correctly when looking at the [CSV on Github](https://github.com/eez-open/modular-psu/blob/69914f9efc4e894c550f1bad077a6b88c76dd525/dcm220/BOM/1-click-bom.csv). 

https://master.staging.kitspace.dev/eez-open/modular-psu/dcm220

PR #683, opened 2024-11-22
Add `--filter` option to import script

AbdulrhmnGhanem:
I find myself want to import specific boards, for example, in #678.


kasbah:
Can we just use prettier? I've already configured my editor for it...

PR #709, opened 2025-01-14
Switch freegeoip endpoint to reallyfreegeoip.org

kasbah:


kasbah:
The endpoint doesn't support CORS.

PR #774, opened 2025-12-30
Remove Open Gamma Project links

NuclearPhoenixx:
Removed several GitHub project links from boards.txt.

### DIFF
--- a/boards.txt
+++ b/boards.txt
@@ -1,13 +1,4 @@
 https://github.com/OpenDosimeter/OpenDosimeter
-https://github.com/OpenGammaProject/MicroFC-TempCo
-https://github.com/OpenGammaProject/AFBR-SiPM-Array-Board
-https://github.com/OpenGammaProject/Tiny-3mm-MicroFC-Carrier-Board
-https://github.com/OpenGammaProject/Tiny-MicroFC-Carrier-Board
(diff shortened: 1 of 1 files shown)


kasbah:
Thanks, I'll go ahead and remove them. Can I ask:

1. Do you want to re-direct the kitspace pages to their respective github repos?
2. Can I ask why you are removing them out of interest? Is it just one thing less to maintain?

NuclearPhoenixx:
Hi there!

1. Sure it would be nice, but you can also just display a 404 or something, I'm not too bothered to be honest.
2. One thing less to maintain is certainly a big one. Every now and then I still get questions from people about the page and how to use it and such. I've also got my fair share of questions in the direction of "how do you use PCB manufacturer xy that you listed on Kitspace", meanwhile I have never used that service and it's just one of the services listed by default kind of thing.

Also, I'm a bit disappointed that there are only two manufacturers listed anymore. When I started I think it was 4? (might be wrong, but it was more than 2) I liked it better when people had more of a choice. Now most people will just use PCBWay because it's cheaper and more comfortable than Aisler, and I have my own link for that, so there is no real benefit anymore.

Hope that helps! Please don't take any of that personal, I think it's a great page, but since I have all of the documentation and links in my README files, I don't need an additional site to present the projects anymore.

